PR #6, opened 2024-02-11
swiss: add support for invariants checking

petermattis:
If the "invariants" build tag is specified, additional invariant checks are performed during operations. These invariant checks are expensive, so they are generally only utilized for testing and debugging.

### DIFF
--- a/map.go
+++ b/map.go
@@ -1,3 +1,17 @@
+// Copyright 2024 The Cockroach Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 // package swiss is a Go implementation of Swiss Tables as described in
 // https://abseil.io/about/design/swisstables. See also:
 // https://faultlore.com/blah/hashbrown-tldr/.
@@ -189,6 +203,7 @@ func New[K comparable, V any](initialCapacity int, options ...option[K, V]) *Map
 		targetCapacity := (uintptr(1) << bits.Len(uint(initialCapacity))) - 1
 		m.resize(targetCapacity)
 	}
+	m.checkInvariants()
 	return m
 }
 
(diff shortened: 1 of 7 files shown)


map.go
```diff
@@ -568,8 +587,10 @@ func (m *Map[K, V]) uncheckedPut(h uintptr, key K, value V) {
 			slot := m.slots.At(i)
 			slot.key = key
 			slot.value = value
+			if *m.ctrls.At(i) == ctrlEmpty {
+				m.growthLeft--
```
petermattis, line 591: This was a tiny bug the invariant checking caught. If the slot was previously deleted we should not have been decrementing `growthLeft`. The impact would be rehashing/resizing slightly more often, though I couldn't measure any effect on the benchmarks.

RaduBerinde (approved):
:lgtm: modulo the tag name

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/6#-:-NqNuHAS79fpwr_4r7Yy:b-drwif9)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on @nvanbenschoten and @petermattis)

___
*[`invariants_off.go` line 15 at r1](https://reviewable.io/reviews/cockroachdb/swiss/6#-NqNtolh4QM2zvxYWqQX:-NqNtolh4QM2zvxYWqQY:bssfusd) ([raw file](https://github.com/cockroachdb/swiss/blob/d056f70b0a0745cc545a3b7eb720ea6eb65931f0/invariants_off.go#L15)):*
> ```Go
> // limitations under the License.
> 
> //go:build !invariants
> ```

We should make this `swiss_invariants` or something similar. Otherwise it will always be on in all Pebble builds with `invariants`..
___
*[`map.go` line 828 at r1](https://reviewable.io/reviews/cockroachdb/swiss/6#-NqNuD2D2t4rEW2__j3f:-NqNuD2D2t4rEW2__j3g:b9znln1) ([raw file](https://github.com/cockroachdb/swiss/blob/d056f70b0a0745cc545a3b7eb720ea6eb65931f0/map.go#L828)):*
> ```Go
> 		}
> 
> 		// For every non-empty slot, verify we can retrieve the key using Get.
> ```

This is great!

petermattis (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/6#-:-NqO3WxxAn8UozdrIOOc:b-814g8k)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on @nvanbenschoten and @RaduBerinde)

___
*[`invariants_off.go` line 15 at r1](https://reviewable.io/reviews/cockroachdb/swiss/6#-NqNtolh4QM2zvxYWqQX:-NqO3OFa65veoK-tM1ey:b-94vdw4) ([raw file](https://github.com/cockroachdb/swiss/blob/d056f70b0a0745cc545a3b7eb720ea6eb65931f0/invariants_off.go#L15)):*

We should make this `swiss_invariants` or something similar. Otherwise it will always be on in all Pebble builds with `invariants`..
</blockquote></details>

Good point. That would make the Pebble tests with `invariants` enabled really slow. Done.

PR #8, opened 2024-02-11
minor comments and readability improvements

RaduBerinde:
Add more comments about some bit operations.

Rename and document the function that converts bytes for rehashing in place. Fix minor omission in the corresponding test.

PR #9, opened 2024-02-11
follow-up to #8

RaduBerinde:
Sorry, I had not pushed these changes before merging #8.


RaduBerinde:
Actually, never mind, I will post a follow-up PR with more comment additions.

PR #10, opened 2024-02-11
tbitset comments and improvements

RaduBerinde:
This commit clarifies that `bitset` represents a set of slots within a group and abstracts usage around it without focusing on the representation. We also move the leading/trailing calculation as methods and add tests.

### DIFF
--- a/map.go
+++ b/map.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// package swiss is a Go implementation of Swiss Tables as described in
+// Package swiss is a Go implementation of Swiss Tables as described in
 // https://abseil.io/about/design/swisstables. See also:
 // https://faultlore.com/blah/hashbrown-tldr/.
 //
@@ -250,8 +250,8 @@ func (m *Map[K, V]) Put(key K, value V) {
 		}
 
 		for match != 0 {
-			bit := match.next()
-			i := seq.offsetAt(bit)
+			slotIdx := match.first()
+			i := seq.offsetAt(slotIdx)
 			if debug {
 				fmt.Printf("put(checking): index=%d  key=%v\n", i, m.slots.At(i).key)
 			}
@@ -264,7 +264,7 @@ func (m *Map[K, V]) Put(key K, value V) {
 				m.checkInvariants()
 				return
 			}
-			match = match.clear(bit)
+			match = match.remove(slotIdx)
 		}
 
 		match = g.matchEmpty()
@@ -335,8 +335,8 @@ func (m *Map[K, V]) Get(key K) (value V, ok bool) {
 		}
 
 		for match != 0 {
-			bit := match.next()
-			i := seq.offsetAt(bit)
+			slotIdx := match.first()
+			i := seq.offsetAt(slotIdx)
 			if debug {
 				fmt.Printf("get(checking): index=%d  key=%v\n", i, m.slots.At(i).key)
 			}
@@ -344,7 +344,7 @@ func (m *Map[K, V]) Get(key K) (value V, ok bool) {
 			if key == slot.key {
 				return slot.value, true
 			}
-			match = match.clear(bit)
+			match = match.remove(slotIdx)
 		}
 
 		match = g.matchEmpty()
@@ -387,8 +387,8 @@ func (m *Map[K, V]) Delete(key K) {
 		}
 
 		for match != 0 {
-			bit := match.next()
-			i := seq.offsetAt(bit)
+			slotIdx := match.first()
+			i := seq.offsetAt(slotIdx)
 			if debug {
 				fmt.Printf("delete(checking): index=%d  key=%v\n", i, m.slots.At(i).key)
 			}
@@ -426,7 +426,7 @@ func (m *Map[K, V]) Delete(key K) {
 				m.checkInvariants()
 				return
 			}
-			match = match.clear(bit)
+			match = match.remove(slotIdx)
 		}
 
 		match = g.matchEmpty()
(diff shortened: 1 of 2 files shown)


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/10)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @petermattis)

___
*[`map.go` line 903 at r1](https://reviewable.io/reviews/cockroachdb/swiss/10#-NqObVjGCE8cN1KxRJvD:-NqObVjGCE8cN1KxRJvE:b-8cwiw1) ([raw file](https://github.com/cockroachdb/swiss/blob/a4fd1ba262b9c5d9fee6a45aa9425de94230f53a/map.go#L903)):*
> ```Go
> }
> 
> // Returns the maximal number of contiguous slots at the beginning of the group
> ```

Maybe I should only have `first()` and `last()` here and do something like `if (7 - before.last()) + after.first()` where we use them?  We would be doing `7 - (7 - LeadingZeros)` but that could get optimized out (?)

petermattis (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/10#-:-NqOjKXW7_GBZCD-HyJi:bnfp4nl)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`map.go` line 903 at r1](https://reviewable.io/reviews/cockroachdb/swiss/10#-NqObVjGCE8cN1KxRJvD:-NqOi4RA-4MrUaCINSom:ba026hb) ([raw file](https://github.com/cockroachdb/swiss/blob/a4fd1ba262b9c5d9fee6a45aa9425de94230f53a/map.go#L903)):*

Maybe I should only have `first()` and `last()` here and do something like `if (7 - before.last()) + after.first()` where we use them?  We would be doing `7 - (7 - LeadingZeros)` but that could get optimized out (?)
</blockquote></details>

Definitely more symmetric to have only `first()` and `last()`. I'm not understanding the `7 - (7 - LeadingZeros)` bit. Wouldn't you do `emptyBefore.last() + emptyAfter.first() < groupSize`?
___
*[`map_test.go` line 187 at r2](https://reviewable.io/reviews/cockroachdb/swiss/10#-NqOirE_AzedqdjXwWM0:-NqOirE_AzedqdjXwWM1:b-5m4wa7) ([raw file](https://github.com/cockroachdb/swiss/blob/cdfd93863830a06ffa4bfdad4ab3c204bb112a84/map_test.go#L187)):*
> ```Go
> 		end   uintptr
> 	}{
> 		{ctrls: []ctrl{0, ctrlEmpty, 0, 0, ctrlEmpty, ctrlEmpty, 0, 0}, start: 1, end: 2},
> ```

Nit: I personally find these more readable if each line is the same length. That's why I used a uint64 in one of the other tests. Another alternative is to specify these as hex strings and then do some unsafe casting below to get a `[]ctrl`. The specification would look like `ctrls: "\x00\x80\x00\x00\x80\x80\x00\x00"`.

RaduBerinde:
*[`map.go` line 903 at r1](https://reviewable.io/reviews/cockroachdb/swiss/10#-NqObVjGCE8cN1KxRJvD:-NqP0xbn-BqS3H6MpNrO:b-2gmkvt) ([raw file](https://github.com/cockroachdb/swiss/blob/a4fd1ba262b9c5d9fee6a45aa9425de94230f53a/map.go#L903)):*

Definitely more symmetric to have only `first()` and `last()`. I'm not understanding the `7 - (7 - LeadingZeros)` bit. Wouldn't you do `emptyBefore.last() + emptyAfter.first() < groupSize`?
</blockquote></details>

The index of the last group would be 7 - LeadingZeros/8

RaduBerinde:
*[`map.go` line 903 at r1](https://reviewable.io/reviews/cockroachdb/swiss/10#-NqObVjGCE8cN1KxRJvD:-NqP1BqtEoJE_MlK2Scr:becjf37) ([raw file](https://github.com/cockroachdb/swiss/blob/a4fd1ba262b9c5d9fee6a45aa9425de94230f53a/map.go#L903)):*

The index of the last group would be 7 - LeadingZeros/8
</blockquote></details>

*last slot

RaduBerinde:
*[`map.go` line 903 at r1](https://reviewable.io/reviews/cockroachdb/swiss/10#-NqObVjGCE8cN1KxRJvD:-NqP5udfB3RRwKfEIxC0:b-bfjl8c) ([raw file](https://github.com/cockroachdb/swiss/blob/a4fd1ba262b9c5d9fee6a45aa9425de94230f53a/map.go#L903)):*

*last slot
</blockquote></details>

We'd want something like reverseFirst = first() on the "reversed" set if we called ReverseBytes64.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/10)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @petermattis)

___
*[`map_test.go` line 187 at r2](https://reviewable.io/reviews/cockroachdb/swiss/10#-NqOirE_AzedqdjXwWM0:-NqQAHrJ4gpn6qU42rn3:b-p2qul5) ([raw file](https://github.com/cockroachdb/swiss/blob/cdfd93863830a06ffa4bfdad4ab3c204bb112a84/map_test.go#L187)):*

Nit: I personally find these more readable if each line is the same length. That's why I used a uint64 in one of the other tests. Another alternative is to specify these as hex strings and then do some unsafe casting below to get a `[]ctrl`. The specification would look like `ctrls: "\x00\x80\x00\x00\x80\x80\x00\x00"`.
</blockquote></details>

Done. I chose to represent the bitset directly as a string.

petermattis (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/10)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[`map_test.go` line 187 at r2](https://reviewable.io/reviews/cockroachdb/swiss/10#-NqOirE_AzedqdjXwWM0:-NqRwNxu-YIWM0Nb-PTm:b-15jgmu) ([raw file](https://github.com/cockroachdb/swiss/blob/cdfd93863830a06ffa4bfdad4ab3c204bb112a84/map_test.go#L187)):*

Done. I chose to represent the bitset directly as a string.
</blockquote></details>

That's even cleaner.

PR #11, opened 2024-02-11
swiss: implement extendible hashing

petermattis:
Add extendible hashing on top of the base Swiss Tables design. When there is
only a single bucket, performance is essentially unchanged. The differences
below are likely just noise. The benefit of extendible hashing is to reduce
the maximum operation time when resizing needs to take place. With extendible
hashing resizing is incremental as we can split a single bucket at a time.
With a max bucket capacity of 4095 (the default) splitting of a bucket takes
in the 100us range. Resizing a bucket has a similar cost, but since the max
bucket size is unbound without extendible hashing it is possible to get up to
10s of milliseconds for the resize operation (e.g. for a million element map).
See TestResizeVsSplit for evidence of this.

```
name                                              old time/op      new time/op      delta
StringMap/avgLoad,n=10/swissMap/Get-10                8.40ns ± 0%      8.30ns ± 0%   -1.09%  (p=0.000 n=10+8)
StringMap/avgLoad,n=83/swissMap/Get-10                8.64ns ± 4%      9.12ns ±15%     ~     (p=0.278 n=9+10)
StringMap/avgLoad,n=671/swissMap/Get-10               9.00ns ± 4%      8.91ns ± 3%     ~     (p=0.247 n=10+10)
StringMap/avgLoad,n=5375/swissMap/Get-10              9.35ns ± 1%      9.28ns ± 1%     ~     (p=0.075 n=10+10)
StringMap/avgLoad,n=86015/swissMap/Get-10             10.9ns ± 1%      10.9ns ± 0%   -0.65%  (p=0.000 n=10+9)
Int64Map/avgLoad,n=10/swissMap/Get-10                 4.85ns ± 1%      4.78ns ± 1%   -1.49%  (p=0.000 n=10+10)
Int64Map/avgLoad,n=83/swissMap/Get-10                 5.23ns ±15%      4.88ns ± 9%   -6.77%  (p=0.013 n=10+9)
Int64Map/avgLoad,n=671/swissMap/Get-10                5.42ns ± 3%      5.28ns ± 5%     ~     (p=0.076 n=9+10)
Int64Map/avgLoad,n=5375/swissMap/Get-10               5.68ns ± 1%      5.60ns ± 2%   -1.46%  (p=0.011 n=10+10)
Int64Map/avgLoad,n=86015/swissMap/Get-10              6.88ns ± 0%      6.67ns ± 0%   -3.12%  (p=0.000 n=10+10)
StringMap/avgLoad,n=10/swissMap/PutDelete-10          25.8ns ±32%      27.0ns ±21%     ~     (p=0.218 n=10+10)
StringMap/avgLoad,n=83/swissMap/PutDelete-10          31.8ns ±10%      31.0ns ±12%     ~     (p=0.393 n=10+10)
StringMap/avgLoad,n=671/swissMap/PutDelete-10         34.1ns ± 4%      34.6ns ± 7%     ~     (p=0.436 n=10+10)
StringMap/avgLoad,n=5375/swissMap/PutDelete-10        35.3ns ± 1%      36.1ns ± 4%   +2.03%  (p=0.005 n=10+10)
StringMap/avgLoad,n=86015/swissMap/PutDelete-10       39.5ns ± 1%      40.1ns ± 1%   +1.50%  (p=0.000 n=9+10)
Int64Map/avgLoad,n=10/swissMap/PutDelete-10           19.4ns ±15%      16.9ns ±23%  -12.91%  (p=0.035 n=9+10)
Int64Map/avgLoad,n=83/swissMap/PutDelete-10           21.3ns ± 6%      20.3ns ± 8%   -4.62%  (p=0.041 n=9+10)
Int64Map/avgLoad,n=671/swissMap/PutDelete-10          24.7ns ± 9%      24.5ns ± 7%     ~     (p=0.739 n=10+10)
Int64Map/avgLoad,n=5375/swissMap/PutDelete-10         27.7ns ± 4%      27.5ns ± 2%     ~     (p=0.517 n=10+10)
Int64Map/avgLoad,n=86015/swissMap/PutDelete-10        32.0ns ± 1%      31.5ns ± 1%   -1.76%  (p=0.000 n=10+10)
```

When configured with a maximum bucket capacity of 4095 performance degrades
somewhat when there are multiple buckets:

```
name                                              old time/op      new time/op      delta
StringMap/avgLoad,n=10/swissMap/Get-10                8.30ns ± 0%      8.31ns ± 0%       ~     (p=0.263 n=8+9)
StringMap/avgLoad,n=83/swissMap/Get-10                9.12ns ±15%      8.86ns ±10%       ~     (p=0.564 n=10+10)
StringMap/avgLoad,n=671/swissMap/Get-10               8.91ns ± 3%      8.92ns ± 3%       ~     (p=0.853 n=10+10)
StringMap/avgLoad,n=5375/swissMap/Get-10              9.28ns ± 1%     11.37ns ± 1%    +22.54%  (p=0.000 n=10+9)
StringMap/avgLoad,n=86015/swissMap/Get-10             10.9ns ± 0%      12.5ns ± 1%    +14.90%  (p=0.000 n=9+9)
Int64Map/avgLoad,n=10/swissMap/Get-10                 4.78ns ± 1%      4.78ns ± 1%       ~     (p=0.704 n=10+9)
Int64Map/avgLoad,n=83/swissMap/Get-10                 4.88ns ± 9%      5.26ns ±17%     +7.71%  (p=0.043 n=9+10)
Int64Map/avgLoad,n=671/swissMap/Get-10                5.28ns ± 5%      5.39ns ± 7%       ~     (p=0.353 n=10+10)
Int64Map/avgLoad,n=5375/swissMap/Get-10               5.60ns ± 2%      6.67ns ± 2%    +19.23%  (p=0.000 n=10+9)
Int64Map/avgLoad,n=86015/swissMap/Get-10              6.67ns ± 0%      8.05ns ± 0%    +20.67%  (p=0.000 n=10+8)
StringMap/avgLoad,n=10/swissMap/PutDelete-10          27.0ns ±21%      28.0ns ±21%       ~     (p=0.497 n=10+9)
StringMap/avgLoad,n=83/swissMap/PutDelete-10          31.0ns ±12%      33.9ns ± 8%     +9.52%  (p=0.003 n=10+10)
StringMap/avgLoad,n=671/swissMap/PutDelete-10         34.6ns ± 7%      35.2ns ± 5%       ~     (p=0.280 n=10+10)
StringMap/avgLoad,n=5375/swissMap/PutDelete-10        36.1ns ± 4%      41.2ns ± 3%    +14.34%  (p=0.000 n=10+10)
StringMap/avgLoad,n=86015/swissMap/PutDelete-10       40.1ns ± 1%      45.7ns ± 1%    +14.01%  (p=0.000 n=10+9)
Int64Map/avgLoad,n=10/swissMap/PutDelete-10           16.9ns ±23%      16.5ns ±13%       ~     (p=0.797 n=10+9)
Int64Map/avgLoad,n=83/swissMap/PutDelete-10           20.3ns ± 8%      20.4ns ± 9%       ~     (p=0.912 n=10+10)
Int64Map/avgLoad,n=671/swissMap/PutDelete-10          24.5ns ± 7%      24.8ns ± 7%       ~     (p=0.353 n=10+10)
Int64Map/avgLoad,n=5375/swissMap/PutDelete-10         27.5ns ± 2%      32.2ns ± 3%    +17.06%  (p=0.000 n=10+10)
Int64Map/avgLoad,n=86015/swissMap/PutDelete-10        31.5ns ± 1%      35.6ns ± 1%    +13.22%  (p=0.000 n=10+8)
```

The size of this degradation is somewhat surprising as locating a bucket
involves a right shift of the hash and indexing into an array of pointers.
Even with this performance degradation, `swiss.Map` significantly outperforms
Go's builtin map:

```
name                                         old time/op  new time/op  delta
StringMap/avgLoad,n=10/Map/Get-10            10.2ns ±17%   8.3ns ± 0%  -18.62%  (p=0.000 n=10+9)
StringMap/avgLoad,n=83/Map/Get-10            10.9ns ± 9%   8.9ns ±10%  -18.97%  (p=0.000 n=10+10)
StringMap/avgLoad,n=671/Map/Get-10           15.4ns ± 5%   8.9ns ± 3%  -42.07%  (p=0.000 n=10+10)
StringMap/avgLoad,n=5375/Map/Get-10          25.8ns ± 1%  11.4ns ± 1%  -56.01%  (p=0.000 n=10+9)
StringMap/avgLoad,n=86015/Map/Get-10         30.2ns ± 1%  12.5ns ± 1%  -58.68%  (p=0.000 n=9+9)
Int64Map/avgLoad,n=10/Map/Get-10             5.00ns ± 0%  4.78ns ± 1%   -4.33%  (p=0.000 n=9+9)
Int64Map/avgLoad,n=83/Map/Get-10             5.19ns ± 2%  5.26ns ±17%     ~     (p=0.353 n=10+10)
Int64Map/avgLoad,n=671/Map/Get-10            6.37ns ±10%  5.39ns ± 7%  -15.39%  (p=0.000 n=10+10)
Int64Map/avgLoad,n=5375/Map/Get-10           17.9ns ± 2%   6.7ns ± 2%  -62.80%  (p=0.000 n=9+9)
Int64Map/avgLoad,n=86015/Map/Get-10          23.8ns ± 0%   8.0ns ± 0%  -66.12%  (p=0.000 n=10+8)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,47 +8,49 @@ implementation in the Rust standard library.
 A `swiss.Map[K,V]` maps keys of type `K` to values of type `V`, similar to
 Go's builtin `map[K]V` type. The primary advantage of `swiss.Map` over Go's
 builtin map is performance. `swiss.Map` has similar or slightly better
-performance Go's builtin map for small map sizes, and significantly better
-performance at large map sizes.
+performance than Go's builtin map for small map sizes, and significantly
+better performance at large map sizes.
 
 ```
 name                                         old time/op  new time/op  delta
-StringMap/avgLoad,n=10/Map/Get-10            9.46ns ± 4%  8.43ns ± 1%  -10.89%  (p=0.000 n=10+9)
-StringMap/avgLoad,n=83/Map/Get-10            10.9ns ± 7%   8.9ns ±12%  -18.45%  (p=0.000 n=10+10)
-StringMap/avgLoad,n=671/Map/Get-10           15.4ns ± 3%   9.1ns ± 3%  -40.98%  (p=0.000 n=10+10)
-StringMap/avgLoad,n=5375/Map/Get-10          25.8ns ± 1%   9.3ns ± 1%  -63.83%  (p=0.000 n=10+9)
-StringMap/avgLoad,n=86015/Map/Get-10         30.4ns ± 1%  10.8ns ± 1%  -64.49%  (p=0.000 n=9+9)
-Int64Map/avgLoad,n=10/Map/Get-10             5.05ns ± 2%  4.87ns ± 1%   -3.60%  (p=0.000 n=10+10)
-Int64Map/avgLoad,n=83/Map/Get-10             5.27ns ± 5%  5.29ns ±12%     ~     (p=0.912 n=10+10)
-Int64Map/avgLoad,n=671/Map/Get-10            6.14ns ± 4%  5.35ns ± 3%  -12.85%  (p=0.000 n=10+10)
-Int64Map/avgLoad,n=5375/Map/Get-10           18.4ns ± 4%   5.7ns ± 2%  -68.94%  (p=0.000 n=10+10)
-Int64Map/avgLoad,n=86015/Map/Get-10          23.9ns ± 0%   6.9ns ± 0%  -71.35%  (p=0.000 n=10+8)
+StringMap/avgLoad,n=10/Map/Get-10            10.2ns ±17%   8.3ns ± 0%  -18.62%  (p=0.000 n=10+9)
+StringMap/avgLoad,n=83/Map/Get-10            10.9ns ± 9%   8.9ns ±10%  -18.97%  (p=0.000 n=10+10)
+StringMap/avgLoad,n=671/Map/Get-10           15.4ns ± 5%   8.9ns ± 3%  -42.07%  (p=0.000 n=10+10)
+StringMap/avgLoad,n=5375/Map/Get-10          25.8ns ± 1%  11.4ns ± 1%  -56.01%  (p=0.000 n=10+9)
+StringMap/avgLoad,n=86015/Map/Get-10         30.2ns ± 1%  12.5ns ± 1%  -58.68%  (p=0.000 n=9+9)
+Int64Map/avgLoad,n=10/Map/Get-10             5.00ns ± 0%  4.78ns ± 1%   -4.33%  (p=0.000 n=9+9)
+Int64Map/avgLoad,n=83/Map/Get-10             5.19ns ± 2%  5.26ns ±17%     ~     (p=0.353 n=10+10)
+Int64Map/avgLoad,n=671/Map/Get-10            6.37ns ±10%  5.39ns ± 7%  -15.39%  (p=0.000 n=10+10)
+Int64Map/avgLoad,n=5375/Map/Get-10           17.9ns ± 2%   6.7ns ± 2%  -62.80%  (p=0.000 n=9+9)
+Int64Map/avgLoad,n=86015/Map/Get-10          23.8ns ± 0%   8.0ns ± 0%  -66.12%  (p=0.000 n=10+8)
 
 name                                         old time/op  new time/op  delta
-StringMap/avgLoad,n=10/Map/PutDelete-10      25.4ns ± 6%  23.7ns ± 8%   -6.43%  (p=0.004 n=10+10)
-StringMap/avgLoad,n=83/Map/PutDelete-10      31.4ns ± 7%  24.3ns ±12%  -22.66%  (p=0.000 n=10+10)
-StringMap/avgLoad,n=671/Map/PutDelete-10     45.4ns ± 3%  24.9ns ± 4%  -45.21%  (p=0.000 n=10+10)
-StringMap/avgLoad,n=5375/Map/PutDelete-10    56.7ns ± 1%  24.7ns ± 2%  -56.44%  (p=0.000 n=10+10)
-StringMap/avgLoad,n=86015/Map/PutDelete-10   60.8ns ± 1%  31.6ns ± 2%  -48.03%  (p=0.000 n=9+9)
-Int64Map/avgLoad,n=10/Map/PutDelete-10       18.0ns ± 3%  17.1ns ±34%     ~     (p=0.095 n=9+10)
-Int64Map/avgLoad,n=83/Map/PutDelete-10       19.8ns ± 3%  14.6ns ±12%  -26.11%  (p=0.000 n=9+9)
-Int64Map/avgLoad,n=671/Map/PutDelete-10      27.2ns ± 3%  15.2ns ± 6%  -44.02%  (p=0.000 n=10+10)
-Int64Map/avgLoad,n=5375/Map/PutDelete-10     44.5ns ± 0%  16.9ns ± 3%  -62.10%  (p=0.000 n=7+10)
-Int64Map/avgLoad,n=86015/Map/PutDelete-10    50.8ns ± 0%  21.0ns ± 1%  -58.65%  (p=0.000 n=10+10)
+StringMap/avgLoad,n=10/Map/PutDelete-10      26.3ns ±13%  28.0ns ±21%     ~     (p=0.133 n=10+9)
+StringMap/avgLoad,n=83/Map/PutDelete-10      30.8ns ± 8%  33.9ns ± 8%  +10.14%  (p=0.000 n=10+10)
+StringMap/avgLoad,n=671/Map/PutDelete-10     45.3ns ± 1%  35.2ns ± 5%  -22.30%  (p=0.000 n=10+10)
+StringMap/avgLoad,n=5375/Map/PutDelete-10    56.5ns ± 1%  41.2ns ± 3%  -27.03%  (p=0.000 n=10+10)
+StringMap/avgLoad,n=86015/Map/PutDelete-10   60.4ns ± 0%  45.7ns ± 1%  -24.24%  (p=0.000 n=10+9)
+Int64Map/avgLoad,n=10/Map/PutDelete-10       18.1ns ± 6%  16.5ns ±13%   -8.84%  (p=0.002 n=9+9)
+Int64Map/avgLoad,n=83/Map/PutDelete-10       19.7ns ±10%  20.4ns ± 9%     ~     (p=0.110 n=10+10)
+Int64Map/avgLoad,n=671/Map/PutDelete-10      27.0ns ± 2%  24.8ns ± 7%   -7.82%  (p=0.000 n=9+10)
+Int64Map/avgLoad,n=5375/Map/PutDelete-10     43.9ns ± 1%  32.2ns ± 3%  -26.63%  (p=0.000 n=10+10)
+Int64Map/avgLoad,n=86015/Map/PutDelete-10    50.4ns ± 0%  35.6ns ± 1%  -29.25%  (p=0.000 n=10+8)
 ```
+
+On top of the base Swiss Tables design, `swiss.Map` adds an [extendible
+hashing](https://en.wikipedia.org/wiki/Extendible_hashing) layer in order to
+enable incremental resizing of large maps.
 
 ## Caveats
 
-- Resizing a `swiss.Map` is done for the whole table rather than the
-  incremental resizing performed by Go's builtin map.
 - The implementation currently requires a little endian CPU architecture. This
   is not a fundamental limitation of the implementation, merely a choice of
   expediency.
 - Go's builtin map has a fast-path for comparing strings that [share their
   underlying
   storage](https://github.com/golang/go/blob/4a7f3ac8eb4381ea62caa1741eeeec28363245b4/src/runtime/map_faststr.go#L100).
-  This fast-path is feasible because `map[string]T` is specialized which isn't
-  currently possible with Go's generics.
+  This fast-path is feasible because `map[string]T` is specialized for string
+  keys which isn't currently possible with Go's generics.
 
 ## TODO
 
@@ -56,9 +58,7 @@ Int64Map/avgLoad,n=86015/Map/PutDelete-10    50.8ns ± 0%  21.0ns ± 1%  -58.65%
   arm64](https://github.com/abseil/abseil-cpp/commit/6481443560a92d0a3a55a31807de0cd712cd4f88)
   - This appears to be somewhat difficult. Naively implementing the match
     routines in assembly isn't workable as the function call overhead
-    dominates the performance improvement form the SIMD comparisons. The
+    dominates the performance improvement from the SIMD comparisons. The
     Abseil implementation is leveraring gcc/llvm assembly intrinsics which are
     not currently available in Go. In order to take advantage of SIMD we'll
     have to write most/all of the probing loop in assembly.
-- Explore extendible hashing to allow incremental resizing. See
-  https://github.com/golang/go/issues/54766#issuecomment-1233125048
--- a/map.go
+++ b/map.go
@@ -68,6 +68,85 @@
 // of the deleting slot are empty which indicates that the slot was never part
 // of a full group.
 //
+// # Extendible Hashing
+//
+// The Swiss table design has a significant caveat: resizing of the table is
+// done all at once rather than incrementally. This can cause long-tail
+// latency blips in some use cases. To address this caveat, extendible hashing
+// (https://en.wikipedia.org/wiki/Extendible_hashing) is applied on top of the
+// Swiss table foundation. In extendible hashing, there is a top-level
+// directory containing entries pointing to buckets. In swiss.Map each bucket
+// is a Swiss table as described above.
+//
+// The high bits of hash(key) are used to index into the bucket directory
+// which is effectively a trie. The number of bits used is the globalDepth,
+// resulting in 2^globalDepth directory entries. Adjacent entries in the
+// directory are allowed to point to the same bucket which enables resizing to
+// be done incrementally, one bucket at a time. Each bucket has a localDepth
+// which is less than or equal to the globalDepth. If the localDepth for a
+// bucket equals the globalDepth then only a single directory entry points to
+// the bucket. Otherwise, more than one directory entry points to the bucket.
+//
+// The diagram below shows one possible scenario for the directory and
+// buckets. With a globalDepth of 2 the directory contains 4 entries. The
+// first 2 entries point to the same bucket which has a localDepth of 1, while
+// the last 2 entries point to different buckets.
+//
+//	 dir(globalDepth=2)
+//	+----+
+//	| 00 | --\
+//	+----+    +--> bucket[localDepth=1]
+//	| 01 | --/
+//	+----+
+//	| 10 | ------> bucket[localDepth=2]
+//	+----+
+//	| 11 | ------> bucket[localDepth=2]
+//	+----+
+//
+// The index into the directory is "hash(key) >> (64 - globalDepth)".
+//
+// When a bucket gets too large (specified by a configurable threshold) it is
+// split. When a bucket is split its localDepth is incremented. If its
+// localDepth is less than or equal to its globalDepth then the newly split
+// bucket can be installed in the directory. If the bucket's localDepth is
+// greater than the globalDepth then the globalDepth is incremented and the
+// directory is reallocated at twice its current size. In the diagram above,
+// consider what happens if the bucket at dir[3] is split:
+//
+//	 dir(globalDepth=3)
+//	+-----+
+//	| 000 | --\
+//	+-----+    \
+//	| 001 | ----\
+//	+-----+      +--> bucket[localDepth=1]
+//	| 010 | ----/
+//	+-----+    /
+//	| 011 | --/
+//	+-----+
+//	| 100 | --\
+//	+-----+    +----> bucket[localDepth=2]
+//	| 101 | --/
+//	+-----+
+//	| 110 | --------> bucket[localDepth=3]
+//	+-----+
+//	| 111 | --------> bucket[localDepth=3]
+//	+-----+
+//
+// Note that the diagram above is very unlikely with a good hash function as
+// the buckets will tend to fill at a similar rate.
+//
+// The split operation redistributes the records in a bucket into two buckets.
+// This is done by walking over the records in the bucket to be split,
+// computing hash(key) and using localDepth to extract the bit which
+// determines whether to leave the record in the current bucket or to move it
+// to the new bucket.
+//
+// Maps containing only a single bucket are optimized to avoid the directory
+// indexing resulting in performance that is equivalent to a Swiss table
+// without extendible hashing. A single bucket can be guaranteed by
+// configuring a very large bucket size threshold via the
+// WithMaxBucketCapacity option.
+//
 // # Implementation
 //
 // The implementation follows Google's Abseil implementation of Swiss Tables,
@@ -111,6 +190,7 @@ package swiss
 
 import (
 	"fmt"
+	"io"
 	"math/bits"
 	"strings"
 	"unsafe"
@@ -130,6 +210,9 @@ const (
 	bitsetMSB     = 0x8080808080808080
 	bitsetEmpty   = bitsetLSB * uint64(ctrlEmpty)
 	bitsetDeleted = bitsetLSB * uint64(ctrlDeleted)
+
+	minBucketCapacity        uintptr = 7
+	defaultMaxBucketCapacity uintptr = 4095
 )
 
 // Slot holds a key and value.
@@ -138,36 +221,25 @@ type Slot[K comparable, V any] struct {
 	value V
 }
 
-// Map is an unordered map from keys to values with Put, Get, Delete, and All
-// operations. It is inspired by Google's Swiss Tables design as implemented
-// in Abseil's flat_hash_map. By default, a Map[K,V] uses the same hash
-// function as Go's builtin map[K]V, though a different hash function can be
-// specified using the WithHash option.
-//
-// A Map is NOT goroutine-safe.
-type Map[K comparable, V any] struct {
+// bucket implements Google's Swiss Tables hash table design. A Map is
+// composed of 1 or more buckets that are addressed using extendible hashing.
+type bucket[K comparable, V any] struct {
 	// ctrls is capacity+groupSize in length. Ctrls[capacity] is always
 	// ctrlSentinel which is used to stop probe iteration. A copy of the first
 	// groupSize-1 elements of ctrls is mirrored into the remaining slots
 	// which is done so that a probe sequence which picks a value near the end
 	// of ctrls will have valid control bytes to look at.
 	//
-	// When the map is empty, ctrls points to emptyCtrls which will never be
-	// modified and is used to simplify the Put, Get, and Delete code which
+	// When the bucket is empty, ctrls points to emptyCtrls which will never
+	// be modified and is used to simplify the Put, Get, and Delete code which
 	// doesn't have to check for a nil ctrls.
 	ctrls ctrlBytes
 	// slots is capacity in length.
 	slots unsafeSlice[Slot[K, V]]
-	// The hash function to each keys of type K. The hash function is
-	// extracted from the Go runtime's implementation of map[K]struct{}.
-	hash hashFn
-	seed uintptr
-	// The allocator to use for the ctrls and slots slices.
-	allocator Allocator[K, V]
 	// The total number slots (always 2^N-1). The capacity is used as a mask
 	// to quickly compute i%N using a bitwise & operation.
 	capacity uintptr
-	// The number of filled slots (i.e. the number of elements in the map).
+	// The number of filled slots (i.e. the number of elements in the bucket).
 	used int
 	// The number of slots we can still fill without needing to rehash.
 	//
@@ -176,9 +248,57 @@ type Map[K comparable, V any] struct {
 	// table is filled with tombstones as otherwise probe sequences might get
 	// unacceptably long without triggering a rehash.
 	growthLeft int
+	// localDepth is the number of high bits from hash(key) used to generate
+	// an index for the global directory to locate this bucket. If localDepth
+	// is 0 this bucket is Map.bucket0.
+	localDepth uint
+	// The index of the bucket within Map.dir. If localDepth < globalDepth
+	// then this is the index of the first entry in Map.dir which points to
+	// this bucket and the following 1<<(globalDepth-localDepth) entries will
+	// also point to this bucket.
+	index uintptr
 }
 
-// New constructs a new M with the specified initial capacity. If
+// Map is an unordered map from keys to values with Put, Get, Delete, and All
+// operations. Map is inspired by Google's Swiss Tables design as implemented
+// in Abseil's flat_hash_map, combined with extendible hashing. By default, a
+// Map[K,V] uses the same hash function as Go's builtin map[K]V, though a
+// different hash function can be specified using the WithHash option.
+//
+// A Map is NOT goroutine-safe.
+type Map[K comparable, V any] struct {
+	// The hash function to each keys of type K. The hash function is
+	// extracted from the Go runtime's implementation of map[K]struct{}.
+	hash hashFn
+	seed uintptr
+	// The allocator to use for the ctrls and slots slices.
+	allocator Allocator[K, V]
+	// bucket0 is always present and inlined in the Map to avoid a pointer
+	// indirection during the common case that the map contains a single
+	// bucket.
+	bucket0 bucket[K, V]
+	// The directory of buckets.
+	dir unsafeSlice[*bucket[K, V]]
+	// The number of filled slots across all buckets (i.e. the number of
+	// elements in the map).
+	used int
+	// globalShift is the number of bits to right shift a hash value to
+	// generate an index for the global directory. As a special case, if
+	// globalShift==0 then bucket0 is used and the directory is not accessed.
+	// Note that globalShift==(64-globalDepth). globalShift is used rather
+	// than globalDepth because the shifting is the more common operation than
+	// needing to compare globalDepth to a bucket's localDepth.
+	globalShift uint
+	// The maximum capacity a bucket is allowed to grow to before it will be
+	// split.
+	maxBucketCapacity uintptr
+}
+
+func normalizeCapacity(capacity uintptr) uintptr {
+	return (uintptr(1) << bits.Len64(uint64(capacity)-1)) - 1
+}
+
+// New constructs a new Map with the specified initial capacity. If
 // initialCapacity is 0 the map will start out with zero capacity and will
 // grow on the first insert. The zero value for an M is not usable.
 func New[K comparable, V any](initialCapacity int, options ...option[K, V]) *Map[K, V] {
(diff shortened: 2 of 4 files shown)


RaduBerinde:
I will have to think about it a bot.

I had a different idea. We have the old and new map and maintain an invariant that all hashes (mod old size) below X are in the new map and the ones above are in the new map. Lookup is no slower because we know in which map to look.

Before every insert we advance X by 4 (moving up to 4 elements to the new map). We insert the new element in either map depending on X. If we start this process before the map is 3/4 full we won't overflow the old map.

I don't know how much of a concern resizing really is in practice in CRDB usecased.

petermattis:
Another pro for the extendible hashing approach is that if we make the bucket size threshold configurable we can configure a very large bucket size in order to fall back to the current behavior.

RaduBerinde (commented):
I read it and it makes a lot of sense. It's a little unfortunate that it adds more complexity into the existing code. It would be nice if it could be separated out more; ideally the current variant would still exist and work on its own.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/11#-:-NqSiUAC0Z_PA6Bx9ePq:b-mfz5dp)* status: 0 of 3 files reviewed, all discussions resolved

petermattis (commented):
> It would be nice if it could be separated out more; ideally the current variant would still exist and work on its own.

I agree with this sentiment, though it might be at odds with getting the most performance. We'd want to compute `hash(key)` only once, so we'd have to expose interfaces to pass in the hash value along with the key. I worry that the overhead of restructuring in this way might add additional function calls that impact performance. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/11#-:-NqSl8dyDL_d5oOvibsX:b-ozi5f0)* status: 0 of 3 files reviewed, all discussions resolved

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/11)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @petermattis)

___
*[`map.go` line 295 at r3](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqYTQW7808n3qRaPadH:-NqYTQW7808n3qRaPadI:b-fp7doa) ([raw file](https://github.com/cockroachdb/swiss/blob/5eef4f8170fb05fdfe2cf566d29cd483c3406f6f/map.go#L295)):*
> ```Go
> 
> func normalizeCapacity(capacity int) uintptr {
> 	// If the user specified a power of 2, just return that value minus 1
> ```

[nit] isn't this the same as using `capacity - 1` in the formula below?

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/11)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`map.go` line 295 at r3](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqYTQW7808n3qRaPadH:-NqYbWr62XhEhND1b7gc:bgprlch) ([raw file](https://github.com/cockroachdb/swiss/blob/5eef4f8170fb05fdfe2cf566d29cd483c3406f6f/map.go#L295)):*

[nit] isn't this the same as using `capacity - 1` in the formula below?
</blockquote></details>

Good point. Btw, I'm going to be taking a stab at fleshing out the implementation of extendible hashing sketched out here.

petermattis:
This is ready for a looksie. I broke a test which is easy to fix, and need to add some more tests. Perf is unchanged if we have only a single bucket. If we have more than 1 bucket perf declines a bit, though still much better than runtime map. Here, old is 1 bucket, and new is more than 1 bucket:

```
name                                        old time/op      new time/op      delta
StringMap/avgLoad,n=10/swissMap/Get-10          8.33ns ± 1%      8.36ns ± 1%     +0.37%  (p=0.019 n=8+10)
StringMap/maxLoad,n=13/swissMap/Get-10          8.33ns ± 0%      8.33ns ± 0%       ~     (p=1.000 n=8+8)
StringMap/minLoad,n=14/swissMap/Get-10          8.59ns ± 1%      8.34ns ± 1%     -2.99%  (p=0.000 n=8+8)
StringMap/avgLoad,n=83/swissMap/Get-10          8.89ns ± 8%      8.77ns ±10%       ~     (p=0.615 n=10+10)
StringMap/maxLoad,n=111/swissMap/Get-10         10.1ns ± 9%      10.1ns ± 2%       ~     (p=0.983 n=10+8)
StringMap/minLoad,n=112/swissMap/Get-10         8.54ns ± 0%      8.35ns ± 2%     -2.25%  (p=0.000 n=9+9)
StringMap/avgLoad,n=671/swissMap/Get-10         9.17ns ± 4%      8.94ns ± 2%     -2.51%  (p=0.011 n=10+10)
StringMap/maxLoad,n=895/swissMap/Get-10         10.3ns ± 3%      10.3ns ± 4%       ~     (p=0.986 n=10+10)
StringMap/minLoad,n=896/swissMap/Get-10         8.43ns ± 1%      8.49ns ± 1%       ~     (p=0.142 n=9+9)
StringMap/avgLoad,n=5375/swissMap/Get-10        9.42ns ± 6%     10.82ns ± 1%    +14.92%  (p=0.000 n=10+9)
StringMap/maxLoad,n=7167/swissMap/Get-10        11.1ns ± 1%      10.9ns ± 1%     -1.74%  (p=0.000 n=9+10)
StringMap/minLoad,n=7168/swissMap/Get-10        8.78ns ± 1%     10.88ns ± 0%    +23.91%  (p=0.000 n=9+9)
StringMap/avgLoad,n=86015/swissMap/Get-10       11.1ns ± 1%      15.4ns ± 6%    +39.28%  (p=0.000 n=9+9)
StringMap/maxLoad,n=114687/swissMap/Get-10      13.9ns ± 1%      17.0ns ± 9%    +22.20%  (p=0.000 n=10+9)
StringMap/minLoad,n=114688/swissMap/Get-10      14.4ns ± 3%      16.8ns ± 3%    +16.38%  (p=0.000 n=10+10)

name                                        old buckets      new buckets      delta
StringMap/avgLoad,n=10/swissMap/Get-10            1.00 ± 0%        1.00 ± 0%       ~     (all equal)
StringMap/maxLoad,n=13/swissMap/Get-10            1.00 ± 0%        1.00 ± 0%       ~     (all equal)
StringMap/minLoad,n=14/swissMap/Get-10            1.00 ± 0%        1.00 ± 0%       ~     (all equal)
StringMap/avgLoad,n=83/swissMap/Get-10            1.00 ± 0%        1.00 ± 0%       ~     (all equal)
StringMap/maxLoad,n=111/swissMap/Get-10           1.00 ± 0%        1.00 ± 0%       ~     (all equal)
StringMap/minLoad,n=112/swissMap/Get-10           1.00 ± 0%        1.00 ± 0%       ~     (all equal)
StringMap/avgLoad,n=671/swissMap/Get-10           1.00 ± 0%        1.00 ± 0%       ~     (all equal)
StringMap/maxLoad,n=895/swissMap/Get-10           1.00 ± 0%        1.00 ± 0%       ~     (all equal)
StringMap/minLoad,n=896/swissMap/Get-10           1.00 ± 0%        1.00 ± 0%       ~     (all equal)
StringMap/avgLoad,n=5375/swissMap/Get-10          1.00 ± 0%        4.00 ± 0%   +300.00%  (p=0.000 n=10+10)
StringMap/maxLoad,n=7167/swissMap/Get-10          1.00 ± 0%        4.00 ± 0%   +300.00%  (p=0.000 n=10+10)
StringMap/minLoad,n=7168/swissMap/Get-10          1.00 ± 0%        4.00 ± 0%   +300.00%  (p=0.000 n=10+10)
StringMap/avgLoad,n=86015/swissMap/Get-10         1.00 ± 0%       64.00 ± 0%  +6300.00%  (p=0.000 n=10+10)
StringMap/maxLoad,n=114687/swissMap/Get-10        1.00 ± 0%       64.00 ± 0%  +6300.00%  (p=0.000 n=10+10)
StringMap/minLoad,n=114688/swissMap/Get-10        1.00 ± 0%       64.00 ± 0%  +6300.00%  (p=0.000 n=10+10)
```

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/11)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`map.go` line 136 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh1U_y8X8cBakOJNBj:-Nqh1U_y8X8cBakOJNBk:b56wxv5) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L136)):*
> ```Go
> //
> // Note that the diagram above is very unlikely with a good hash function as
> // the buckets will tend to fill at a similar rate.
> ```

@RaduBerinde This occurs in practice. Do you have any thoughts on how to smooth out the splitting over time? Is that even worthwhile to do?

RaduBerinde (commented):
This is a very cool technique and the code came out nice.

It would be great to add a metamorphic test that cross-checks against maps with various bucket sizes. Operations should include mutations during iteration.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/11#-:-Nqh1qpcCYD1ye3z_ifL:bjykixg)* status: 0 of 4 files reviewed, 23 unresolved discussions (waiting on @petermattis)

___
*[`map.go` line 136 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh1U_y8X8cBakOJNBj:-Nqh3S9u2VWHmZlcSWlW:b-br1ekp) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L136)):*

@RaduBerinde This occurs in practice. Do you have any thoughts on how to smooth out the splitting over time? Is that even worthwhile to do?
</blockquote></details>

What is the problem exactly?  I don't think it matters if the local depths varies a lot..
___
*[`map.go` line 215 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh6ZE88ZRzuvQsafXw:-Nqh6ZE88ZRzuvQsafXx:b-ik72o0) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L215)):*
> ```Go
> 
> 	minBucketCapacity        uintptr = 7
> 	defaultMaxBucketCapacity uintptr = 4095
> ```

4k seems pretty small IMO, I think we should expect the caller to customize it down when per-op latency is critical (and/or objects are large), and the default can be higher (e.g. 64k or 128k)
___
*[`map.go` line 326 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqgvgHaA0qllIViTaPW:-NqgvgHaA0qllIViTaPX:b41ijyo) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L326)):*
> ```Go
> 			m.maxBucketCapacity = minBucketCapacity
> 		}
> 		m.maxBucketCapacity = normalizeCapacity(m.maxBucketCapacity)
> ```

[nit] could just do `max(m.maxBucketCapacity, minBucketCapacity)`
___
*[`map.go` line 328 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqgvUJgCRQDmiw_SqUa:-NqgvUJgCRQDmiw_SqUb:b-peytys) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L328)):*
> ```Go
> 		m.maxBucketCapacity = normalizeCapacity(m.maxBucketCapacity)
> 	} else {
> 		m.maxBucketCapacity = defaultMaxBucketCapacity
> ```

[nit] We could just move this line up to before applying the options, then we don't need the zero check.
___
*[`map.go` line 344 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqgw9gJ9c8-fpkJCDQx:-Nqgw9gJ9c8-fpkJCDQy:btguh8y) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L344)):*
> ```Go
> 			// size the directory appropriately. We'll size each bucket to
> 			// maxBucketCapacity and create enough buckets to hold
> 			// initialCapacity. Probably easiest to just create one bucket of
> ```

[nit] The "probably easiest" part doesn't apply (right?) Maybe remove or move under a TODO.
___
*[`map.go` line 353 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqgxPlS2ax_yFkLKoyN:-NqgxPlS2ax_yFkLKoyO:br5w008) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L353)):*
> ```Go
> 
> 			i := uintptr(0)
> 			m.dirEntries(func(b *bucket[K, V]) bool {
> ```

This is a very awkward way to initialize, we should just do the loop.

Also, we could allocate all buckets at once (in a slice). And I don't think it matters if we use bucket0 in this case.
___
*[`map.go` line 380 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqgyfuZAp17SJl0jnes:-NqgyfuZAp17SJl0jnet:b-tftmle) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L380)):*
> ```Go
> func (m *Map[K, V]) Close() {
> 	m.buckets(0, func(b *bucket[K, V]) bool {
> 		if b.capacity > 0 {
> ```

[nit] This code should be a method on `bucket`
___
*[`map.go` line 445 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqgymna-ccxGbppgPIz:-Nqgymna-ccxGbppgPJ-:b-hg4ohk) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L445)):*
> ```Go
> 					seq.offset, match, b.ctrls.Slice(seq.offset, seq.offset+groupSize))
> 			}
> 			// Before performing the insertion we may decide the table is getting
> ```

the bucket is getting overcrowded
___
*[`map.go` line 710 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh2bD7CeJrGU94Pjg7:-Nqh2bD7CeJrGU94Pjg8:bs9zzan) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L710)):*
> ```Go
> 	// Map.bucket(), Map.bucketCount(), and bucketStep(). This technique was
> 	// lifted from the Go runtime's runtime/map.go:bucketShift() routine.
> 	ptrSize   = 4 << (^uintptr(0) >> 63)
> ```

[nit] maybe just mention that `ptrSize` is 4 or 8 depending on 32-bit or 64-bit.

By the way, we should add a 32-bit CI target that runs tests with `GOARCH=386` on linux.
___
*[`map.go` line 729 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqgyGIJAYFs3ERLH9i_:-NqgyGIJAYFs3ERLH9ia:b-s5efjx) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L729)):*
> ```Go
> // buckets calls yield sequentially for each bucket in the map. If yield
> // returns false, iteration stops. Offset specifies which which bucket to
> // start iteration at which is used to randomize iteration order.
> ```

[nit] at witch is hard to parse, maybe "start iteration at (used to randomize..)"
___
*[`map.go` line 737 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh-ywbBQ060v74VLZP:-Nqh-ywbBQ060v74VLZQ:blezqxd) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L737)):*
> ```Go
> 
> 	// While iterating the size of the directory can grow if the map is
> 	// mutated. We want to iterate over each bucket once, and if a bucket
> ```

[nit] if the yield function mutates the map
___
*[`map.go` line 754 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh20AU9hOp2kKrIeAD:-Nqh20AU9hOp2kKrIeAE:bwkjkki) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L754)):*
> ```Go
> 		}
> 
> 		i := (b.index + bucketStep(m.globalDepth(), localDepth)) & (m.bucketCount() - 1)
> ```

[nit] Add a comment explaining that `b.index` won't change. Or just save the pre-yield value so there's less to think about.
___
*[`map.go` line 815 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh3GaD6HTL2ZprX-sP:-Nqh3GaD6HTL2ZprX-sQ:b-4n5jzi) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L815)):*
> ```Go
> // growDirectory grows the directory slice to 1<<newGlobalDepth buckets.
> func (m *Map[K, V]) growDirectory(newGlobalDepth uint) {
> 	newDir := makeUnsafeSlice(make([]*bucket[K, V], 1<<newGlobalDepth))
> ```

I'd add an invariants check that `newGlobalDepth` is not unreasonably large
___
*[`map.go` line 933 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqgzsM-0p_rBOcD4bN_:-NqgzsM-0p_rBOcD4bNa:b-obts8x) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L933)):*
> ```Go
> // insertion.
> func (b *bucket[K, V]) uncheckedPut(h uintptr, key K, value V) {
> 	// Given key and its hash hash(key), to insert it, we construct a
> ```

Add an invariants check that growthLeft > 0
___
*[`map.go` line 990 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh-0VN5HOw_iHo4aEr:-Nqh-0VN5HOw_iHo4aEs:b-vswd1u) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L990)):*
> ```Go
> 	// size as the current bucket.
> 	newCapacity := 2*b.capacity + 1
> 	if newCapacity <= m.maxBucketCapacity {
> ```

[nit] I'd invert the condition so that it's consistent with the comment
___
*[`map.go` line 999 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqgxxtG7iB4B4LM0R2R:-NqgxxtG7iB4B4LM0R2S:b-wq6tjb) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L999)):*
> ```Go
> }
> 
> func (b *bucket[K, V]) init(m *Map[K, V], newCapacity uintptr) {
> ```

Maybe this should take `localDepth` as well.
___
*[`map.go` line 1024 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh-5hd5Wnqf_RYoZwn:-Nqh-5hd5Wnqf_RYoZwo:b-9gpq3p) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L1024)):*
> ```Go
> }
> 
> // resize resize the capacity of the table by allocating a bigger array and
> ```

[nit] stutter
___
*[`map.go` line 1057 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqgyzy8DFsnyd2ft1UP:-Nqgyzy8DFsnyd2ft1UQ:btoqc6b) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L1057)):*
> ```Go
> 
> // split divides the entries in a bucket between the receiver and a new bucket
> // of the same size, and the installs the new bucket into the buckets
> ```

[nit] then
___
*[`map.go` line 1065 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh7eUyEEYjoXX9Zg_j:-Nqh7eUyEEYjoXX9Zg_k:bv3v6en) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L1065)):*
> ```Go
> 		index:      b.index,
> 	}
> 	newb.init(m, b.capacity)
> ```

In the degenerate cases we might be wasting a lot of memory, maybe we should let it grow naturally. Ah, in the degenerate case where we move all (or most) of the elements, we'd have the same problem. We may want to consider rebuilding either bucket to a smaller size at the end (something that would rarely happen in practice). Could be a TODO for now
___
*[`map.go` line 1069 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh0UB26ydNHUEOhXaB:-Nqh0UB26ydNHUEOhXaC:b-fq5hhg) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L1069)):*
> ```Go
> 	// Divide the records between the 2 buckets (b and newb). This is done by
> 	// examining the new bit in the hash that will be added to the bucket
> 	// index. If that bit is 0 the record stays in bucket b. If that bit is 1
> ```

I'd mention here that we are relying on the bucket b staying earlier in the directory (for .buckets()).
___
*[`map.go` line 1101 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh7Qc-3zI0TpZL3_aZ:-Nqh7Qc-3zI0TpZL3_a_:bdkssei) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L1101)):*
> ```Go
> 	}
> 
> 	if newb.used == 0 {
> ```

What if we only moved a very small number of elements? We could end up calling split frequently. We should ensure that the buckets have the desired load factor in all cases.
___
*[`map.go` line 1106 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqgzVHgAHfcZomGLDnQ:-NqgzVHgAHfcZomGLDnR:bhm68fs) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L1106)):*
> ```Go
> 		// degenerate hash function (e.g. one that returns a constant in the
> 		// high bits).
> 		m.maxBucketCapacity = 2*m.maxBucketCapacity + 1
> ```

[nit] we should add a debug log here
___
*[`map.go` line 1135 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh-Dt-3LX-q2TbNPGb:-Nqh-Dt-3LX-q2TbNPGc:b-eolrs9) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L1135)):*
> ```Go
> 	}
> 
> 	// Complete the split by increment local depth for the 2 buckets and
> ```

incrementing the local depth

RaduBerinde (commented):
:lgtm: 

Unrelated, but do you know if the go standard `map` resizes down when it gets too empty?

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/11#-:-NqiI3i858ee60U0rRPK:b-2kzikn)* status: 0 of 4 files reviewed, 9 unresolved discussions (waiting on @petermattis)

___
*[`map.go` line 136 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh1U_y8X8cBakOJNBj:-NqiFeeDBwvpU54kIRwG:babb59n) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L136)):*

If your hash function is evenly distributing keys across buckets, then the buckets will all be filling at the same rate and all become full at approximately the same time. What we'll see is that a map will have some number of buckets all being filled up, and then there will be a period of time when they all are becoming full and splitting. Then there will be another long period (presumably twice as long as the previous period) where they are filling up. I don't see how the local depth will vary a lot across buckets unless the hash function is bad.
</blockquote></details>

Oh, so you're worried we'll have this "square wave" in terms of performance? Interesting.. Not sure if there's a good way around that. We will reach a steady state size at some point.
___
*[`map.go` line 215 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh6ZE88ZRzuvQsafXw:-NqiFvR2B45hrbQndGNc:bibqxtd) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L215)):*

Hmm. I can see that argument. I was imagining the opposite. That the caller would increase the default when absolute throughput was critical, and that per-op latency should be the default optimized configuration.
</blockquote></details>

Heh I can see your argument too. I think either way works, we just need to give some more detailed guidance for `New`
___
*[`map.go` line 326 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqgvgHaA0qllIViTaPW:-NqiG7Ko8_Ngt5knLikt:b-k1gsv) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L326)):*

Heh, I tried that originally. `max` was introduced in go1.21. I could make that the minimum. Is it worth supporting go1.20?
</blockquote></details>

Ah, never mind, not worth reducing the compatibility.
___
*[`map.go` line 754 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh20AU9hOp2kKrIeAD:-NqiIquPBv-KxKKuUSQ5:bl1ma6u) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L754)):*

Well, `b.index` can change and we need to use the updated value. I've added a small comment which points to the directory diagrams at the top of this file to aid in understanding.
</blockquote></details>

Ah, yes, makes sense.
___
*[`map.go` line 815 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh3GaD6HTL2ZprX-sP:-NqiIKLv-q8AuIFcQFdQ:bb73djm) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L815)):*

Done. I put the limit at 32. That's 4 billion directory entries. Should be enough. Do you have a better suggestion?
</blockquote></details>

SGTM
___
*[`map.go` line 999 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqgxxtG7iB4B4LM0R2R:-NqiI_m3Ampb2DL9hntZ:b-1d9pew) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L999)):*

Can you expand on what you're thinking would help here? During `resize` the `localDepth` isn't changing, and during `split` we don't change the `localDepth` until we've discovered that the split bucket is actually worth keeping and we don't have a degenerate hash function.
</blockquote></details>

Just felt weird in the initialization path to call `init()` and then having to set `localDepth` separately. Not a big deal either way
___
*[`map.go` line 1065 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh7eUyEEYjoXX9Zg_j:-NqiH6Sm-tvyn4Ho2gEe:bl6viu) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L1065)):*

Can you elaborate on the memory wastage? In the degenerate case we don't keep one of the buckets and we don't grow the directory. So we did allocate some memory for the split, and then we resize the bucket we're keeping. I don't see a way to do that better.
</blockquote></details>

I'm thinking of a case where say 25% of the elements go to one bucket. Probably not worth worrying about it though.
___
*[`map.go` line 1101 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh7Qc-3zI0TpZL3_aZ:-NqiGc-78s9FQc_LnUDT:b-p5z7yt) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L1101)):*

What are you imagining for a check? Certainly possible for a hash function to be bad, but not degenerately bad. A bad hash function can mess up any hash table implementation. How defensive should we be?
</blockquote></details>

I would just change the check to be something along the lines of:
`if b.used > b.capacity * 7 / 8 { resize() }`
___
*[`map.go` line 301 at r9](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqiG4L_9M00QYu04aqx:-NqiG4L_9M00QYu04aqy:b-qjeave) ([raw file](https://github.com/cockroachdb/swiss/blob/b2c1f3089d3ec2ff68df2d290a5f7d4d45910102/map.go#L301)):*
> ```Go
> }
> 
> // New constructs a new M with the specified initial capacity. If
> ```

[nit] Map
___
*[`map.go` line 740 at r9](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqiJ7Dd-QaionimNaD6:-NqiJ7Dd-QaionimNaD7:b-pdgy1c) ([raw file](https://github.com/cockroachdb/swiss/blob/b2c1f3089d3ec2ff68df2d290a5f7d4d45910102/map.go#L740)):*
> ```Go
> 	startBucket := *m.dir.At(offset & (m.bucketCount() - 1))
> 	for b := startBucket; ; {
> 		localDepth := b.localDepth
> ```

[nit] maybe name this `originalLocalDepth` to make it even more clear.

petermattis (commented):
> Unrelated, but do you know if the go standard `map` resizes down when it gets too empty?

I don't know. I've looked at the map code and haven't seen anything about resizing down, but I haven't been looking for that specifically. I'll check.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/11#-:-NqiUfCC-vlnQwiOmtvN:b-gq4glw)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @RaduBerinde)

___
*[`map.go` line 136 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh1U_y8X8cBakOJNBj:-NqiUm2o5_h_dTeAJj4n:btih660) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L136)):*

Oh, so you're worried we'll have this "square wave" in terms of performance? Interesting.. Not sure if there's a good way around that. We will reach a steady state size at some point.
</blockquote></details>

Yes, exactly.
___
*[`map.go` line 215 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh6ZE88ZRzuvQsafXw:-NqiUwUQ21UwPqOpbPK_:b-9dembb) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L215)):*

Heh I can see your argument too. I think either way works, we just need to give some more detailed guidance for `New`
</blockquote></details>

I'll see about providing some additional guidance in a follow-on PR.
___
*[`map.go` line 754 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh20AU9hOp2kKrIeAD:-NqiVdF1CjqKaPDBFiN2:bdnean2) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L754)):*

Ah, yes, makes sense.
</blockquote></details>

I will admit this was quite the difficult loop to write. I'm not surprised it was a difficult loop to read. I might add another diagram here to explain what is going on.
___
*[`map.go` line 999 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqgxxtG7iB4B4LM0R2R:-NqiW7mbCVJpM6aWkVSQ:b-5dxvfa) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L999)):*

Just felt weird in the initialization path to call `init()` and then having to set `localDepth` separately. Not a big deal either way
</blockquote></details>

Ack. I'm going to leave as-is.
___
*[`map.go` line 1065 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh7eUyEEYjoXX9Zg_j:-NqiWAzF9ZLeJut5916O:b-uk3hz8) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L1065)):*

I'm thinking of a case where say 25% of the elements go to one bucket. Probably not worth worrying about it though.
</blockquote></details>

Got it. I'm not sure what to do about this. Hash tables need good hash functions are else there are problematic cases. I think the implementation should really only protect against truly degenerate cases.
___
*[`map.go` line 1101 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh7Qc-3zI0TpZL3_aZ:-NqiZu-HCMMXXkstv2nG:b-l4fpfr) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L1101)):*

I would just change the check to be something along the lines of:
`if b.used > b.capacity * 7 / 8 { resize() }`
</blockquote></details>

Ok. Done.
___
*[`map.go` line 301 at r9](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqiG4L_9M00QYu04aqx:-NqiV1BF3V4zlSfnbuuQ:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/b2c1f3089d3ec2ff68df2d290a5f7d4d45910102/map.go#L301)):*

[nit] Map
</blockquote></details>

Done.
___
*[`map.go` line 740 at r9](https://reviewable.io/reviews/cockroachdb/swiss/11#-NqiJ7Dd-QaionimNaD6:-NqiVc2YB5djiFfewGmg:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/b2c1f3089d3ec2ff68df2d290a5f7d4d45910102/map.go#L740)):*

[nit] maybe name this `originalLocalDepth` to make it even more clear.
</blockquote></details>

Done.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/11#-:-NqicXmq89eBTi_ItOLW:bnfp4nl)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @petermattis)

___
*[`map.go` line 754 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh20AU9hOp2kKrIeAD:-NqibtziA6vZ8yrALQAw:bj9rfn9) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L754)):*

I will admit this was quite the difficult loop to write. I'm not surprised it was a difficult loop to read. I might add another diagram here to explain what is going on.
</blockquote></details>

Each bucket contains a range of hashes. If we had a helper function that gave us the range, we could assert (in invariants mode) that the previous bucket pre-yield end of range is the next bucket's start.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/11)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[`map.go` line 754 at r7](https://reviewable.io/reviews/cockroachdb/swiss/11#-Nqh20AU9hOp2kKrIeAD:-NqiwBNb7jxpWk3rWNdJ:b-ozlhck) ([raw file](https://github.com/cockroachdb/swiss/blob/a53d57de0b491b63c58b4a41b74d07f260fec60e/map.go#L754)):*

Each bucket contains a range of hashes. If we had a helper function that gave us the range, we could assert (in invariants mode) that the previous bucket pre-yield end of range is the next bucket's start.
</blockquote></details>

Yeah, that's a good idea to add some invariant checking here. (Though I feel relatively confident it is right at this point).

RaduBerinde:
I was thinking - could we make `New` return a `Map` instead of `*Map` so the struct itself can live on the stack (or be part of larger structs)?

I think we would need some changes to the implementation here so we don't make the pointer to the Map escape. For one, we can't point to `&bucket0` anymore from the directory, but we don't need to, we can just shallow copy the bucket to a new bucket. We would also need to rethink `dirEntries` (we could just special case the globalDepth==0 case wherever we use it).

petermattis:
> I was thinking - could we make New return a Map instead of *Map so the struct itself can live on the stack (or be part of larger structs)?
>
> I think we would need some changes to the implementation here so we don't make the pointer to the Map escape. For one, we can't point to &bucket0 anymore from the directory, but we don't need to, we can just shallow copy the bucket to a new bucket. We would also need to rethink dirEntries (we could just special case the globalDepth==0 case wherever we use it).

I did add support for a `Map.Init` method in https://github.com/cockroachdb/swiss/pull/17 which allows a `Map` to be embedded by value. That doesn't allow the struct to live on the stack by itself, but I'm not sure if that is a common case worth optimizing. Seems like the more common case is embedding a `Map` in another struct. Perhaps I should add a `Make` function as well, similar to `New` but it returns a `Map` instead of a `*Map`. I suspect it is better to just wait for a bit of usage experience.

RaduBerinde:
Thinking more about it, it's not safe to copy a Map by value so we shouldn't encourage that pattern with a Make function.

petermattis:
Good point. We can and should protect against that.

PR #13, opened 2024-02-12
fix Mac OS X CI

petermattis:
The OS is macos-latest, not darwin-latest. Not sure where I grabbed the latter from.


RaduBerinde (approved):
I was wondering why they seemed stuck :)

PR #15, opened 2024-02-12
rename Map.DebugString to Map.GoString

petermattis:
This allows us to use the "%#v" format specifier rather than explicitly invoking `Map.DebugString`.

### DIFF
--- a/map.go
+++ b/map.go
@@ -798,12 +798,12 @@ func (m *Map[K, V]) checkInvariants() {
 				ci := *m.ctrls.At(i)
 				cj := *m.ctrls.At(j)
 				if ci != cj {
-					panic(fmt.Sprintf("invariant failed: ctrl(%d)=%02x != ctrl(%d)=%02x\n%s", i, ci, j, cj, m.DebugString()))
+					panic(fmt.Sprintf("invariant failed: ctrl(%d)=%02x != ctrl(%d)=%02x\n%#v", i, ci, j, cj, m))
 				}
 			}
 			// Verify the sentinel is good.
 			if c := *m.ctrls.At(m.capacity); c != ctrlSentinel {
-				panic(fmt.Sprintf("invariant failed: ctrl(%d): expected sentinel, but found %02x\n%s", m.capacity, c, m.DebugString()))
+				panic(fmt.Sprintf("invariant failed: ctrl(%d): expected sentinel, but found %02x\n%#v", m.capacity, c, m))
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/15)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`map.go` line 865 at r1](https://reviewable.io/reviews/cockroachdb/swiss/15#-NqTr2BxA3D5XPTyRNBZ:-NqTr2BxA3D5XPTyRNB_:b-gaaitg) ([raw file](https://github.com/cockroachdb/swiss/blob/a5ef6699cce96fb004b4d538833f4dd0120d1e5d/map.go#L865)):*
> ```Go
> 				fmt.Fprintf(&buf, "  %4d: %v [ctrl=%02x h2=%02x] \n", i, s.key, c, h2(h))
> 			} else {
> 				fmt.Fprintf(&buf, "  %4d: [ctrl=%02x]\n", i, c)
> ```

This didn't seem useful when visualizing the contents of a `Map`, though I'm still wondering if there is a way to better display the mirrored control bytes. Omitting them is one possibility, but I'd like to include them since I think this output will typically be used in a panic or error context.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/15#-:-NqUUiRYDUcRT6pZUjos:bnfp4nl)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @petermattis)

___
*[`map.go` line 848 at r1](https://reviewable.io/reviews/cockroachdb/swiss/15#-NqUUcwbDZXKWzsvYyNF:-NqUUcwbDZXKWzsvYyNG:b-2k00b4) ([raw file](https://github.com/cockroachdb/swiss/blob/ff0ccd79ee298ccf56c47de1992ffdd6764f4059/map.go#L848)):*
> ```Go
> }
> 
> func (m *Map[K, V]) GoString() string {
> ```

[nit] Worth mentioning here that this is the string that is shown when a `%#v` verb is used with fmt functions.
___
*[`map.go` line 865 at r1](https://reviewable.io/reviews/cockroachdb/swiss/15#-NqTr2BxA3D5XPTyRNBZ:-NqUTvvjBfwsNtfm8vVE:b-odfsmv) ([raw file](https://github.com/cockroachdb/swiss/blob/a5ef6699cce96fb004b4d538833f4dd0120d1e5d/map.go#L865)):*

This didn't seem useful when visualizing the contents of a `Map`, though I'm still wondering if there is a way to better display the mirrored control bytes. Omitting them is one possibility, but I'd like to include them since I think this output will typically be used in a panic or error context.
</blockquote></details>

I don't see a problem with including them given that we have a separate line per entry. Maybe include the hex in all cases `%4d %02x: ` followed by `empty/deleted/sentinel` or the value and hash. I'd show the entire hash value.

petermattis (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/15#-:-NqUYLEo7sXH4g-MlDPQ:b-814g8k)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`map.go` line 848 at r1](https://reviewable.io/reviews/cockroachdb/swiss/15#-NqUUcwbDZXKWzsvYyNF:-NqUXNcZ-Zrdk9_JvjyQ:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/ff0ccd79ee298ccf56c47de1992ffdd6764f4059/map.go#L848)):*

[nit] Worth mentioning here that this is the string that is shown when a `%#v` verb is used with fmt functions.
</blockquote></details>

Done.
___
*[`map.go` line 865 at r1](https://reviewable.io/reviews/cockroachdb/swiss/15#-NqTr2BxA3D5XPTyRNBZ:-NqUXOSt7wPFY4OUfH9d:bnhf3ed) ([raw file](https://github.com/cockroachdb/swiss/blob/a5ef6699cce96fb004b4d538833f4dd0120d1e5d/map.go#L865)):*

I don't see a problem with including them given that we have a separate line per entry. Maybe include the hex in all cases `%4d %02x: ` followed by `empty/deleted/sentinel` or the value and hash. I'd show the entire hash value.
</blockquote></details>

Actually, the hash value thing was only an attempt at debugging something very specific that turned out not to be about mismatched hash/ctrl values anyways. I've removed it and added display of `[<key>:<value>]` for full slots.

PR #17, opened 2024-02-16
swiss: add benchmarks from github.com/zhangyunhao116/gomapbench

petermattis:
These benchmarks cover a wider variety of scenarios than what was previously being tested.


petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/17)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`README.md` line 38 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnadhGA7vjvifdnEyG:-NqnadhGA7vjvifdnEyH:beoryzw) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L38)):*
> ```Markdown
> MapIter/Int/8192-16                  90.6µs ± 1%    71.0µs ± 1%    -21.65%  (p=0.000 n=9+10)
> MapIter/Int/65536-16                  724µs ± 1%     663µs ± 1%     -8.43%  (p=0.000 n=9+9)
> MapGetHit/Int64/6-16                 4.93ns ± 0%    6.54ns ± 0%    +32.72%  (p=0.000 n=10+8)
> ```

This is mentioned in the caveats section: Go's builtin map has an optimization for maps that fit in a single bucket that avoids calling the hash function for int32, int64, and string keys. Not sure if there is anything that can be replicated in the implementation here, or if this is even worth worrying about. It affects all of the 6 key variants.
___
*[`README.md` line 123 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnaVmfEE3FHENTjM1Q:-NqnaVmfEE3FHENTjM1R:b-5syip1) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L123)):*
> ```Markdown
> MapGetMiss/String/65536-16           23.8ns ± 2%    21.4ns ±10%    -10.00%  (p=0.000 n=10+10)
> MapPutGrow/Int64/6-16                83.5ns ± 0%   210.4ns ± 1%   +151.86%  (p=0.000 n=10+10)
> MapPutGrow/Int64/12-16                422ns ± 1%     602ns ± 1%    +42.46%  (p=0.000 n=10+10)
> ```

I mention this in the caveats section: Go's builtin map uses fewer allocations at smaller map sizes which seems to be the performance difference here. I didn't include the `-benchmem` output in the README.md, but you can see this in the allocs/op data:

```
MapPutGrow/Int64/6-16                  0.00           3.00 ± 0%      +Inf%  (p=0.000 n=10+10)
MapPutGrow/Int64/12-16                 1.00 ± 0%      5.00 ± 0%   +400.00%  (p=0.000 n=10+10)
MapPutGrow/Int64/18-16                 3.00 ± 0%      7.00 ± 0%   +133.33%  (p=0.000 n=10+10)
MapPutGrow/Int64/24-16                 4.00 ± 0%      7.00 ± 0%    +75.00%  (p=0.000 n=10+10)
MapPutGrow/Int64/30-16                 6.00 ± 0%      9.00 ± 0%    +50.00%  (p=0.000 n=10+10)
MapPutGrow/Int64/64-16                 12.0 ± 0%      11.0 ± 0%     -8.33%  (p=0.000 n=10+10)
MapPutGrow/Int64/128-16                19.0 ± 0%      13.0 ± 0%    -31.58%  (p=0.000 n=10+10)
MapPutGrow/Int64/256-16                27.0 ± 0%      15.0 ± 0%    -44.44%  (p=0.000 n=10+10)
MapPutGrow/Int64/512-16                42.0 ± 0%      17.0 ± 0%    -59.52%  (p=0.000 n=10+10)
MapPutGrow/Int64/1024-16               64.0 ± 0%      19.0 ± 0%    -70.31%  (p=0.000 n=10+10)
MapPutGrow/Int64/2048-16               99.4 ± 1%      21.0 ± 0%    -78.87%  (p=0.000 n=10+10)
MapPutGrow/Int64/4096-16                161 ± 0%        25 ± 0%    -84.50%  (p=0.000 n=10+10)
MapPutGrow/Int64/8192-16                274 ± 0%        32 ± 0%    -88.32%  (p=0.002 n=8+10)
MapPutGrow/Int64/65536-16             2.35k ± 0%     0.12k ± 0%    -94.93%  (p=0.000 n=10+10)
```

The `MapPutGrow/Int64/6` case is super interesting. How are they getting away with 0 allocations? I've seen some references to maps being allocated on the stack, but I haven't found anything in the code about the bucket being allocated on the stack.

I think it would be possible to eliminate 1 allocate in the `swiss.Map` case by bundling together allocation of the controls and slots for a fixed set of "small" capacity sizes (e.g. 7, 15, 31, 63). Not sure this is worth optimizing.
___
*[`README.md` line 276 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnaCrTA-8dkunNHiwX:-NqnaCrTA-8dkunNHiwY:b9l7vg1) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L276)):*
> ```Markdown
> MapPutDelete/Int32/8192-16           54.3ns ± 1%    37.2ns ± 1%    -31.48%  (p=0.000 n=9+10)
> MapPutDelete/Int32/65536-16          60.5ns ± 1%    43.0ns ± 1%    -28.88%  (p=0.000 n=10+10)
> MapPutDelete/String/6-16             32.5ns ± 1%    40.5ns ± 4%    +24.53%  (p=0.000 n=10+10)
> ```

I switched to gathering benchmark numbers on on `n1-standard-16` roachprod machine because they now take 1+ hour to run. These `MapPutDelete/String` numbers are interesting because they differ from what I'm seeing on my M1 laptop.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/17)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @petermattis)

___
*[`README.md` line 20 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnpZjb-zNNYffWhM8w:-NqnpZjb-zNNYffWhM8x:b-x1aisr) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L20)):*
> ```Markdown
> 
> <details>
> <summary>Performance: Go's builtin map (old) vs swiss.Map (new)</summary>
> ```

[nit] mention that it's on a `n1-standard-16` (Intel, linux) machine
___
*[`README.md` line 38 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnadhGA7vjvifdnEyG:-NqnlEWQ9WlEMPTPv4c0:b-5y0cl1) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L38)):*

This is mentioned in the caveats section: Go's builtin map has an optimization for maps that fit in a single bucket that avoids calling the hash function for int32, int64, and string keys. Not sure if there is anything that can be replicated in the implementation here, or if this is even worth worrying about. It affects all of the 6 key variants.
</blockquote></details>

Couldn't we do something similar along the lines of using hash 0 if capacity is <= 8? Or do we think this would be very detrimental if the types are expensive to compare?

By the way, we should probably grow from 0 directly to 8 in the first add into an empty bucket (so we have at least a group).
___
*[`README.md` line 123 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnaVmfEE3FHENTjM1Q:-NqnoZnb5O21OW0cJmrC:b-av4lol) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L123)):*

I mention this in the caveats section: Go's builtin map uses fewer allocations at smaller map sizes which seems to be the performance difference here. I didn't include the `-benchmem` output in the README.md, but you can see this in the allocs/op data:

```
MapPutGrow/Int64/6-16                  0.00           3.00 ± 0%      +Inf%  (p=0.000 n=10+10)
MapPutGrow/Int64/12-16                 1.00 ± 0%      5.00 ± 0%   +400.00%  (p=0.000 n=10+10)
MapPutGrow/Int64/18-16                 3.00 ± 0%      7.00 ± 0%   +133.33%  (p=0.000 n=10+10)
MapPutGrow/Int64/24-16                 4.00 ± 0%      7.00 ± 0%    +75.00%  (p=0.000 n=10+10)
MapPutGrow/Int64/30-16                 6.00 ± 0%      9.00 ± 0%    +50.00%  (p=0.000 n=10+10)
MapPutGrow/Int64/64-16                 12.0 ± 0%      11.0 ± 0%     -8.33%  (p=0.000 n=10+10)
MapPutGrow/Int64/128-16                19.0 ± 0%      13.0 ± 0%    -31.58%  (p=0.000 n=10+10)
MapPutGrow/Int64/256-16                27.0 ± 0%      15.0 ± 0%    -44.44%  (p=0.000 n=10+10)
MapPutGrow/Int64/512-16                42.0 ± 0%      17.0 ± 0%    -59.52%  (p=0.000 n=10+10)
MapPutGrow/Int64/1024-16               64.0 ± 0%      19.0 ± 0%    -70.31%  (p=0.000 n=10+10)
MapPutGrow/Int64/2048-16               99.4 ± 1%      21.0 ± 0%    -78.87%  (p=0.000 n=10+10)
MapPutGrow/Int64/4096-16                161 ± 0%        25 ± 0%    -84.50%  (p=0.000 n=10+10)
MapPutGrow/Int64/8192-16                274 ± 0%        32 ± 0%    -88.32%  (p=0.002 n=8+10)
MapPutGrow/Int64/65536-16             2.35k ± 0%     0.12k ± 0%    -94.93%  (p=0.000 n=10+10)
```

The `MapPutGrow/Int64/6` case is super interesting. How are they getting away with 0 allocations? I've seen some references to maps being allocated on the stack, but I haven't found anything in the code about the bucket being allocated on the stack.

I think it would be possible to eliminate 1 allocate in the `swiss.Map` case by bundling together allocation of the controls and slots for a fixed set of "small" capacity sizes (e.g. 7, 15, 31, 63). Not sure this is worth optimizing.
</blockquote></details>

I found this comment:
```
// makemap implements Go map creation for make(map[k]v, hint).
// If the compiler has determined that the map or the first bucket
// can be created on the stack, h and/or bucket may be non-nil.
// If h != nil, the map can be created directly in h.
// If h.buckets != nil, bucket pointed to can be used as the first bucket.
func makemap(t *maptype, hint int, h *hmap) *hmap {
```

And this one in `walkMakeMap`:
```
		// Allocate one bucket pointed to by hmap.buckets on stack if hint
		// is not larger than BUCKETSIZE. In case hint is larger than
		// BUCKETSIZE runtime.makemap will allocate the buckets on the heap.
		// Maximum key and elem size is 128 bytes, larger objects
		// are stored with an indirection. So max bucket size is 2048+eps.
		if !ir.IsConst(hint, constant.Int) ||
```

Yeah I'm not sure it's worth optimizing either..
___
*[`README.md` line 276 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnaCrTA-8dkunNHiwX:-NqnpKhxCZx-B2x7o0g5:b-va0ir) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L276)):*

I switched to gathering benchmark numbers on on `n1-standard-16` roachprod machine because they now take 1+ hour to run. These `MapPutDelete/String` numbers are interesting because they differ from what I'm seeing on my M1 laptop.
</blockquote></details>

It's strange that they suddenly get much better between 2048-8192

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/17)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @RaduBerinde)

___
*[`README.md` line 20 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnpZjb-zNNYffWhM8w:-NqnwPuR29qQ-rqpJwou:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L20)):*

[nit] mention that it's on a `n1-standard-16` (Intel, linux) machine
</blockquote></details>

Done.
___
*[`README.md` line 38 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnadhGA7vjvifdnEyG:-Nqnvauk6pftZW2S-iAE:b-r5hhzg) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L38)):*
Yeah, I'm worried that the linear search could be detrimental if types are expensive to compare. For string Go does a complicated comparison of the first 4 bytes, the last 4 bytes, and only if those matches do they compare the middle bytes. 

> By the way, we should probably grow from 0 directly to 8 in the first add into an empty bucket (so we have at least a group).

This is already done by init(). The smallest bucket capacity we create is 7 which has 15 control bytes and 7 slots.
___
*[`README.md` line 123 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnaVmfEE3FHENTjM1Q:-Nqnw5VLBeKds7tO7qS-:b-lybas1) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L123)):*
Ah-ha! Thanks for tracking that down.

> I think it would be possible to eliminate 1 allocate in the `swiss.Map` case by bundling together allocation of the controls and slots for a fixed set of "small" capacity sizes (e.g. 7, 15, 31, 63). Not sure this is worth optimizing.

This was relatively easy to test and moved the needle somewhat for the smallest map sizes on my laptop but had only a very tiny effect on Linux.
___
*[`README.md` line 276 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnaCrTA-8dkunNHiwX:-NqnwGFrCw-qP1B9-cdJ:bkk8jf5) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L276)):*

It's strange that they suddenly get much better between 2048-8192
</blockquote></details>

Yeah. I need to profile this. Something weird is going on.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/17)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`map.go` line 401 at r3](https://reviewable.io/reviews/cockroachdb/swiss/17#-Nqokjmd2MjUIV5ZPSNL:-Nqokjmd2MjUIV5ZPSNM:bbf9bnk) ([raw file](https://github.com/cockroachdb/swiss/blob/abf5390f89e7eb5958c8d306dccbcf0c3e898e73/map.go#L401)):*
> ```Go
> 			// If there is room left to grow in the table and we're at the
> 			// start of the probe sequence we can just insert the new entry.
> 			if b.growthLeft > 0 && seq.offset == startOffset {
> ```

These two fast-paths get hit with reasonable frequency. Unless we're near capacity, we usually find an empty slot at the start of the probe sequence. Similarly, if we can't find an empty slot we often find a deleted slot in our probe sequence that we can overwrite. The latter fast-path comes from Abseil which I had somehow missed earlier. The former fast-path is my own observation (at least I didn't see where Abseil was doing something similar).
___
*[`README.md` line 276 at r1](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqnaCrTA-8dkunNHiwX:-NqolJjP63uym72LCqOJ:b-cxtopz) ([raw file](https://github.com/cockroachdb/swiss/blob/63cab15c3b6b2a9eb707b4958899f1bc9c0ca36a/README.md#L276)):*

Yeah. I need to profile this. Something weird is going on.
</blockquote></details>

The two fast-paths I added improve the situation, but there is still something weird going on with this benchmark on the GCE machine. On my laptop the numbers here look sensical. Some of the int32 and int64 PutDelete numbers are also strange.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/17#-:-NqtCDCR1CbMkeZFJ5vi:bnfp4nl)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @petermattis)

___
*[`map.go` line 399 at r4](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqtAfTMEuhmAV0tBwww:-NqtAfTMEuhmAV0tBwwx:b-vvw32p) ([raw file](https://github.com/cockroachdb/swiss/blob/a1e025a8bebc469b146380099822d4c72705b1dd/map.go#L399)):*
> ```Go
> 			// sequence.
> 
> 			// If there is room left to grow in the table and we're at the
> ```

in the bucket
___
*[`map.go` line 423 at r4](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqtBQ8w8krDLnt_S6JU:-NqtBQ8w8krDLnt_S6JV:bl91dnp) ([raw file](https://github.com/cockroachdb/swiss/blob/a1e025a8bebc469b146380099822d4c72705b1dd/map.go#L423)):*
> ```Go
> 					i := seq.offsetAt(match.first())
> 					// If there is room left to grow in the table or the slot
> 					// is deleted (and thus we're overwriting it andnot
> ```

and not

petermattis (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/17#-:-Nqtbosn5nssxaZ3PZh4:b-814g8k)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`map.go` line 399 at r4](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqtAfTMEuhmAV0tBwww:-NqtbjZv9FOMcjSwB1xO:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/a1e025a8bebc469b146380099822d4c72705b1dd/map.go#L399)):*

in the bucket
</blockquote></details>

Done.
___
*[`map.go` line 423 at r4](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqtBQ8w8krDLnt_S6JU:-Nqtbla0-HG-nOWY_18H:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/a1e025a8bebc469b146380099822d4c72705b1dd/map.go#L423)):*

and not
</blockquote></details>

Done.
___
*[`options.go` line 119 at r5](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqrqptJ7d6kO-kZh_U3:-NqrqptJ7d6kO-kZh_U4:bjrkml) ([raw file](https://github.com/cockroachdb/swiss/blob/e4a2f5b16f5d2dd7d64f3b918404c212568e0652/options.go#L119)):*
> ```Go
> 		a := &bucketAlloc31[K, V]{}
> 		return a.ctrls[:ctrls], a.slots[:slots]
> 	}
> ```

@RaduBerinde This was the allocation optimization I mentioned in another comment which has a decent impact when layered on top of the previous change. I'm somewhat on the fence about this. Sort of neat that it can be encapsulated fully in the Allocator interface, but perhaps it is an over-optimization. I'm not sure it does any harm, though. And there is an argument to be made that this should go in `defaultAllocator`. No real reason not to. Thoughts?

RaduBerinde:
*[`options.go` line 119 at r5](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqrqptJ7d6kO-kZh_U3:-Nqu2wWF5X7AW0vbPuiX:b-lssyeu) ([raw file](https://github.com/cockroachdb/swiss/blob/e4a2f5b16f5d2dd7d64f3b918404c212568e0652/options.go#L119)):*

@RaduBerinde This was the allocation optimization I mentioned in another comment which has a decent impact when layered on top of the previous change. I'm somewhat on the fence about this. Sort of neat that it can be encapsulated fully in the Allocator interface, but perhaps it is an over-optimization. I'm not sure it does any harm, though. And there is an argument to be made that this should go in `defaultAllocator`. No real reason not to. Thoughts?
</blockquote></details>

Could we just embed say 32 or 63 ctrl bytes in each bucket and use those when possible?

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/17)* status: 0 of 6 files reviewed, 2 unresolved discussions

___
*[`options.go` line 119 at r5](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqrqptJ7d6kO-kZh_U3:-Nqu6q860lxY4cdn2wsf:bjzpef6) ([raw file](https://github.com/cockroachdb/swiss/blob/e4a2f5b16f5d2dd7d64f3b918404c212568e0652/options.go#L119)):*

Could we just embed say 32 or 63 ctrl bytes in each bucket and use those when possible?
</blockquote></details>

Wouldn't that waste quite a bit of space? The ctrl bytes need to be contiguous, so those 31 or 63 ctrl bytes would only be usable at those specific bucket sizes.

RaduBerinde:
*[`map.go` line 401 at r3](https://reviewable.io/reviews/cockroachdb/swiss/17#-Nqokjmd2MjUIV5ZPSNL:-NquEwLp9FYAGaY2CFf6:b-83sjoh) ([raw file](https://github.com/cockroachdb/swiss/blob/abf5390f89e7eb5958c8d306dccbcf0c3e898e73/map.go#L401)):*

These two fast-paths get hit with reasonable frequency. Unless we're near capacity, we usually find an empty slot at the start of the probe sequence. Similarly, if we can't find an empty slot we often find a deleted slot in our probe sequence that we can overwrite. The latter fast-path comes from Abseil which I had somehow missed earlier. The former fast-path is my own observation (at least I didn't see where Abseil was doing something similar).
</blockquote></details>

Nice!

RaduBerinde:
*[`options.go` line 119 at r5](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqrqptJ7d6kO-kZh_U3:-NquEZvH4DlOTNbuU1au:b-9k0nvy) ([raw file](https://github.com/cockroachdb/swiss/blob/e4a2f5b16f5d2dd7d64f3b918404c212568e0652/options.go#L119)):*

Wouldn't that waste quite a bit of space? The ctrl bytes need to be contiguous, so those 31 or 63 ctrl bytes would only be usable at those specific bucket sizes.
</blockquote></details>

Let's say we do 31, then we would waste 32 bytes per bucket, which has (I think) 56 bytes right now.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/17)* status: 0 of 6 files reviewed, 2 unresolved discussions

___
*[`options.go` line 119 at r5](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqrqptJ7d6kO-kZh_U3:-NqwQStMEL48myzhpXpM:b31opgd) ([raw file](https://github.com/cockroachdb/swiss/blob/e4a2f5b16f5d2dd7d64f3b918404c212568e0652/options.go#L119)):*

Let's say we do 31, then we would waste 32 bytes per bucket, which has (I think) 56 bytes right now.
</blockquote></details>

We need capacity+groupSize control bytes. The capacities less than 32 are 7, 15, and 31. So we'd need 15, 23 control bytes. Let's say we went with 23 control bytes which would pad out to 24 control bytes. Those inline control bytes would only be usable when the bucket size is 7 or 15, and otherwise would be wasted space. I'm not seeing why this would be better than the approach taken in this PR. Am I missing something about what you're suggesting?

RaduBerinde:
*[`options.go` line 119 at r5](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqrqptJ7d6kO-kZh_U3:-Nqwug6mAGuIhtFzA2_l:bki7x3u) ([raw file](https://github.com/cockroachdb/swiss/blob/e4a2f5b16f5d2dd7d64f3b918404c212568e0652/options.go#L119)):*

We need capacity+groupSize control bytes. The capacities less than 32 are 7, 15, and 31. So we'd need 15, 23 control bytes. Let's say we went with 23 control bytes which would pad out to 24 control bytes. Those inline control bytes would only be usable when the bucket size is 7 or 15, and otherwise would be wasted space. I'm not seeing why this would be better than the approach taken in this PR. Am I missing something about what you're suggesting?
</blockquote></details>

Yeah sorry let's say 24 bytes to support 15 slots. Yes we would waste 24 bytes when we have more than 10-15 elements but by then the space for the slots would dominate anyway.

Not better than your approach, just a simpler way to achieve most of the benefi (since you were debating whether we should keep it).

In any case, whatever we have here should be enabled by default (and not something the user needs to worry about).

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/17)* status: 0 of 6 files reviewed, 2 unresolved discussions

___
*[`options.go` line 119 at r5](https://reviewable.io/reviews/cockroachdb/swiss/17#-NqrqptJ7d6kO-kZh_U3:-NqxHjkT2eafs1hw-O08:bm9xjgz) ([raw file](https://github.com/cockroachdb/swiss/blob/e4a2f5b16f5d2dd7d64f3b918404c212568e0652/options.go#L119)):*

Yeah sorry let's say 24 bytes to support 15 slots. Yes we would waste 24 bytes when we have more than 10-15 elements but by then the space for the slots would dominate anyway.

Not better than your approach, just a simpler way to achieve most of the benefi (since you were debating whether we should keep it).

In any case, whatever we have here should be enabled by default (and not something the user needs to worry about).
</blockquote></details>

Ack on making this enabled by default. I'll stay with the current approach. It wasn't difficult to add and very isolated.

PR #18, opened 2024-02-18
swiss: mark a Map as "noCopy"

petermattis:
It isn't safe to copy a Map by value, so mark it as "noCopy" so that `go vet`
will emit a warning if it is copied.

### DIFF
--- a/map.go
+++ b/map.go
@@ -268,6 +268,7 @@ type Map[K comparable, V any] struct {
 	// The maximum capacity a bucket is allowed to grow to before it will be
 	// split.
 	maxBucketCapacity uintptr
+	_                 noCopy
 }
 
 func normalizeCapacity(capacity uintptr) uintptr {
@@ -1616,3 +1617,16 @@ func (s unsafeSlice[T]) Slice(start, end uintptr) []T {
 func unsafeConvertSlice[Dest any, Src any](s []Src) []Dest {
 	return unsafe.Slice((*Dest)(unsafe.Pointer(unsafe.SliceData(s))), len(s))
 }
+
+// noCopy may be added to structs which must not be copied
+// after the first use.
+//
+// See https://golang.org/issues/8005#issuecomment-190753527
+// for details.
+//
+// Note that it must not be embedded, due to the Lock and Unlock methods.
+type noCopy struct{}
+
+// Lock is a no-op used by -copylocks checker from `go vet`.
+func (*noCopy) Lock()   {}
+func (*noCopy) Unlock() {}


petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/18)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`map.go` line 299 at r1](https://reviewable.io/reviews/cockroachdb/swiss/18#-NqwTIYo1aErxI2lRtjz:-NqwTIYo1aErxI2lRtk-:b-g11n9z) ([raw file](https://github.com/cockroachdb/swiss/blob/81ca8e685db59369699470410208250c978fa99a/map.go#L299)):*
> ```Go
> 	// operation, but because growthLeft == 0 if we try to insert we'll
> 	// immediately rehash and grow.
> 	*m = Map[K, V]{
> ```

@RaduBerinde Do you know why `go vet` isn't complaining about this assignment? I did create a small test case that verifies it will complain if you return a `Map` by value. Perhaps `go vet` knows not to complain if the assignment is happening within a `*Map` method.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/18)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @petermattis)

___
*[`map.go` line 1631 at r1](https://reviewable.io/reviews/cockroachdb/swiss/18#-Nqww9B60MaS0V3UwLDS:-Nqww9B60MaS0V3UwLDT:b9ch3rg) ([raw file](https://github.com/cockroachdb/swiss/blob/81ca8e685db59369699470410208250c978fa99a/map.go#L1631)):*
> ```Go
> 
> // Lock is a no-op used by -copylocks checker from `go vet`.
> func (*noCopy) Lock()   {}
> ```

Should the receiver be non-pointer? Is there a difference?

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/18)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`map.go` line 1631 at r1](https://reviewable.io/reviews/cockroachdb/swiss/18#-Nqww9B60MaS0V3UwLDS:-NqxH_taBVR24pyX4jvT:b-llt2v7) ([raw file](https://github.com/cockroachdb/swiss/blob/81ca8e685db59369699470410208250c978fa99a/map.go#L1631)):*

Should the receiver be non-pointer? Is there a difference?
</blockquote></details>

Not sure there is a difference. This is the way this struct is defined in CRDB (`pkg/util/nocopy.go`) and in the Go runtime.

petermattis:
@RaduBerinde Anything else to do on this PR?

petermattis:
TFTR!

PR #21, opened 2024-02-20
swiss: optimize for number of cache accesses

petermattis:
Switch to interleaving 8 bytes of controls and 8 K/V slots rather than using separate parallel arrays of controls and slots. For very large maps or cold maps (where the map data is not in L2 cache) the interleaving reduces the expected number of cache misses from 2 (controls and slot) to 1 (group). The interleaving does mean that we no longer use unaligned groups which reduces the number of bits of info we're extracting from hash(key), but in practice this doesn't seem to have any impact for a good hash function (the Go runtime hash function is good). In addition to the cache access benefit, this change also means we're always allocating `[]Group[K,V]` instead of separate `[]ctrl` and `[]Slot[K,V]` which reduces the number of memory allocations.

Change `Map.dir` to reference buckets by value instead of by pointer. For large/cold maps, this reduces the number of cache misses to locate a bucket from 2 to 1. This also has a number of other beneficial simplifications because we can now make `Map.dir` always contain all the buckets (it can point to `&m.bucket0` if there is only a single bucket). As part of this change, the `bucket[K,V]` struct was optimized for space, replacing the 64-bit integer fields with 32-bit fields which will restrict the max capacity of a bucket to 4 billion slots. `Sizeof(bucket[K,V]` is now 32 bytes which means 2 buckets can fit in the typical 64-byte cache line.

Added larger benchmark sizes to show the benefits of these changes. Note that these changes were not pure wins vs the previous implementation. In particular, the `MapGetMiss` benchmark regressed on sizes where the controls fit mostly in cache. The comparisons against Go's runtime map continue to look good and no longer look worse at very large map sizes.


petermattis:
@RaduBerinde Most of the changes were fairly mechanical. The difficult changes were primarily in `split` and `growDirectory` as the bucket being split "moves" when the directory is grown. The motivation for this pretty significant change is the handling of large maps which I noticed first when comparing the Pebble Robin-hood map benchmarks which use a 1m entry map.

RaduBerinde (commented):
Very nice. I have to review it in more depth and only have some time here and there this week.

Do you have any idea if we lose anything significant by only using 8-byte aligned groups? I know rust hashbrown made the same choice, and it does make things simpler conceptually, so I wonder why the original implementation had arbitrary groups.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/21#-:-NrBJ4LE8xDUFLlxbRa9:b-49ny9f)* status: 0 of 5 files reviewed, 9 unresolved discussions (waiting on @petermattis)

___
*[`bench_test.go` line 243 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBF6miCbCBays70xJR:-NrBF6miCbCBays70xJS:bywvg06) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/bench_test.go#L243)):*
> ```Go
> 		return true
> 	})
> 	b.ReportMetric(100*float64(fullGroups)/float64(groupsCount), "%full")
> ```

[nit] maybe `fullgrp` or` fullgroups`
___
*[`map.go` line 52 at r1](https://reviewable.io/reviews/cockroachdb/swiss/21#-Nr8ZWQoEa7B_BptRzeS:-Nr8ZWQoEa7B_BptRzeT:b6mejox) ([raw file](https://github.com/cockroachdb/swiss/blob/a295d76caa0b6f9134cdccfd4786cec0ae11733f/map.go#L52)):*
> ```Go
> // for a large or cold map rather than separate accesses for the controls and
> // slots. The mirrored control bytes are no longer needed and we always probe
> // a group's entire set of control bytes.
> ```

[nit] replace "we always probe a group's entire set of control bytes" (which was technically true before) with "and groups no longer start at arbitrary slot index, but only at those that are multiples of 8"
___
*[`map.go` line 337 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBM4j614nGf3rLNDYb:-NrBM4j614nGf3rLNDYc:bxi8318) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L337)):*
> ```Go
> 			// insertion is performed growthLeft==0 will trigger a resize of
> 			// the bucket.
> 			groups: makeUnsafeSlice(unsafeConvertSlice[Group[K, V]](emptyGroup)),
> ```

This is kind of confusing.. I would just make an `emptyCtrls` variable and make a pointer to that into a slice (e.g.` unsafeSlice{ptr: unsafe.Pointer(&emptyCtrls)}`). Going through `Group[string{}, string{}]` is the same technically and only adds confusion.
___
*[`map.go` line 588 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBHYz03vPL8QvxbWTt:-NrBHYz03vPL8QvxbWTu:bl7csgu) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L588)):*
> ```Go
> 				*s = slot[K, V]{}
> 
> 				// Given an offset to delete we simply create a tombstone and
> ```

We can make this a bit more simple and clear:

Only a full group can appear in the middle of a probe sequence (a group with at least one empty slot terminates probing). Once a group becomes full, it stays full until rehashing/resizing. So if the group isn't full now, we can simply remove the element. Otherwise, we create a tombstone to mark the slot as deleted.
___
*[`map.go` line 596 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBIG9S4igkoQfPMw_-:-NrBIG9S4igkoQfPMw_0:b-b6owoa) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L596)):*
> ```Go
> 				// never create an empty slot in a group with no empties which
> 				// this heuristic guarantees we never do).
> 				if g.ctrls.matchEmpty() != 0 {
> ```

Well, this became much simpler :)
___
*[`map.go` line 737 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBILIJAMZDR4vPR--6:-NrBILIJAMZDR4vPR--7:b-7qa48v) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L737)):*
> ```Go
> func (m *Map[K, V]) mutableBucket(h uintptr) *bucket[K, V] {
> 	// NB: It is faster to check for the single bucket case using a
> 	// conditional than to to index into the directory.
> ```

[nit] to
___
*[`map.go` line 742 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBIa7W7c1ENi5Ar80z:-NrBIa7W7c1ENi5Ar81-:b-at6m79) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L742)):*
> ```Go
> 	}
> 	// When shifting by a variable amount the Go compiler inserts overflow
> 	// checks that the shift is less than the maximum allowed (32 or 64).
> ```

We can just use 31 for shiftMask now
___
*[`map.go` line 1483 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBKr7M5HhqMsHD6K2K:-NrBKr7M5HhqMsHD6K2L:bfahb47) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L1483)):*
> ```Go
> type ctrlGroup uint64
> 
> func unsafeCtrlGroup(ctrls []ctrl) *ctrlGroup {
> ```

This is only used in tests, I'd move to `map_test.go` and make it a regular function that builds a `ctrlGroup` using `Set`.
___
*[`options.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBJx9O5_ZzCg1rUbvb:-NrBJx9O5_ZzCg1rUbvc:bh3kzey) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/options.go#L19)):*
> ```Go
> import "unsafe"
> 
> // Option provide an interface to do work on Map while it is being created.
> ```

[nit] "do work on Map" sounds strange as a public API, maybe "an interface for passing configuration parameters for Map initialization".

petermattis (commented):
TFTR!

> Do you have any idea if we lose anything significant by only using 8-byte aligned groups? I know rust hashbrown made the same choice, and it does make things simpler conceptually, so I wonder why the original implementation had arbitrary groups.

As far as I can tell we don't lose anything. Before I embarked on this change I ran an experiment where I changed `h1` to return 8-byte aligned offsets. I couldn't measure a perf change. I agree this makes things conceptually simpler and it makes the implementation simpler. 

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/21#-:-NrCjfVi86caZn57nM03:bj1s8fq)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @RaduBerinde)

___
*[`bench_test.go` line 243 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBF6miCbCBays70xJR:-NrChao3EQwY-Zrsz-HO:b-pfqjvw) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/bench_test.go#L243)):*

[nit] maybe `fullgrp` or` fullgroups`
</blockquote></details>

Thanks for the suggestion. I went with `%fullgrp`. 
___
*[`map.go` line 52 at r1](https://reviewable.io/reviews/cockroachdb/swiss/21#-Nr8ZWQoEa7B_BptRzeS:-NrChr3NAWksnSWx8cKz:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/a295d76caa0b6f9134cdccfd4786cec0ae11733f/map.go#L52)):*

[nit] replace "we always probe a group's entire set of control bytes" (which was technically true before) with "and groups no longer start at arbitrary slot index, but only at those that are multiples of 8"
</blockquote></details>

Done.
___
*[`map.go` line 337 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBM4j614nGf3rLNDYb:-NrChuwn8Civ0MIrFCys:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L337)):*

This is kind of confusing.. I would just make an `emptyCtrls` variable and make a pointer to that into a slice (e.g.` unsafeSlice{ptr: unsafe.Pointer(&emptyCtrls)}`). Going through `Group[string{}, string{}]` is the same technically and only adds confusion.
</blockquote></details>

Done.
___
*[`map.go` line 588 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBHYz03vPL8QvxbWTt:-NrCihc-BnHfdowPh1PT:byedw68) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L588)):*

We can make this a bit more simple and clear:

Only a full group can appear in the middle of a probe sequence (a group with at least one empty slot terminates probing). Once a group becomes full, it stays full until rehashing/resizing. So if the group isn't full now, we can simply remove the element. Otherwise, we create a tombstone to mark the slot as deleted.
</blockquote></details>

Yes, that is much clearer. Done.
___
*[`map.go` line 596 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBIG9S4igkoQfPMw_-:-NrCimf39y-Eb5RTr-zn:b-ve3yvb) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L596)):*

Well, this became much simpler :)
</blockquote></details>

:smile: 
___
*[`map.go` line 737 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBILIJAMZDR4vPR--6:-NrCisTw252UEkTvJ6ns:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L737)):*

[nit] to
</blockquote></details>

Done.
___
*[`map.go` line 742 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBIa7W7c1ENi5Ar80z:-NrCisxF-tF3g1C73aAv:bp7sli8) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L742)):*

We can just use 31 for shiftMask now
</blockquote></details>

Heh, I thought so too until I tried and everything broke. `h` is still a 64-bit `uintptr` and `globalShift` starts at 63 and decreases. 
___
*[`map.go` line 1483 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBKr7M5HhqMsHD6K2K:-NrCjTLj2dVh-VRMJi3h:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L1483)):*

This is only used in tests, I'd move to `map_test.go` and make it a regular function that builds a `ctrlGroup` using `Set`.
</blockquote></details>

Done.
___
*[`options.go` line 19 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBJx9O5_ZzCg1rUbvb:-NrCjdCJ3KLgEftLKoHl:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/options.go#L19)):*

[nit] "do work on Map" sounds strange as a public API, maybe "an interface for passing configuration parameters for Map initialization".
</blockquote></details>

Done.

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/21)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @RaduBerinde)

___
*[`map.go` line 866 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrCl0Em-0VipSd2pAqO:-NrG-cK81_VOfU0i2wSC:b-qu6n94) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L866)):*

Here we can use 31 since we're working on a 32-bit. If we use 63 the compiler can't omit the check (I'm guessing)
</blockquote></details>

Good point. Done.
___
*[`map.go` line 1483 at r2](https://reviewable.io/reviews/cockroachdb/swiss/21#-NrBKr7M5HhqMsHD6K2K:-NrG-zkJ8ZpO3D07mIwK:b1ae97) ([raw file](https://github.com/cockroachdb/swiss/blob/1be9c56f8578286a923d20f6d044c58f597c45c1/map.go#L1483)):*

Done.
</blockquote></details>

Ah, the `TestConvertNonFullToEmptyAndFullToDeleted` informed me why I had this as `unsafeCtrlGroup`. That test expects that it is mutating the slice passed to `unsafeCtrlGroup`. I could adjust the test, but that seems more complicated than just using the `unsafeCtrlGroup` approach.

RaduBerinde (commented):
Do we still need the sentinel? I'm not entirely sure we were using it before either, e.g. in Get we didn't stop if the group had a sentinel (and no other empty slots). I don't quite understand how/why stopping at the sentinel works in the Abseil implementation

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/21#-:-NraQSFH5GnT0ntS3_GD:bc8m2se)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @petermattis)

___
*[`bench_test.go` line 128 at r4](https://reviewable.io/reviews/cockroachdb/swiss/21#-NraMDYa-46ko9qL1S_7:-NraMDYa-46ko9qL1S_8:bwciqc0) ([raw file](https://github.com/cockroachdb/swiss/blob/666532607c5ae2a72222e22ed2fd4b8b9dacd34e/bench_test.go#L128)):*
> ```Go
> 		4096,
> 		8192,
> 		1 << 15,
> ```

[nit] Do we need all these sizes?
___
*[`map.go` line 235 at r4](https://reviewable.io/reviews/cockroachdb/swiss/21#-NraQFTT-QSvB-FbdjWu:-NraQFTT-QSvB-FbdjWv:b9tu3om) ([raw file](https://github.com/cockroachdb/swiss/blob/666532607c5ae2a72222e22ed2fd4b8b9dacd34e/map.go#L235)):*
> ```Go
> 	groups unsafeSlice[Group[K, V]]
> 	// The mask used to compute the index of the group within the bucket. The
> 	// groupMask only changes when a bucket is resized.
> ```

[nit] would be good to explain what it is exactly
___
*[`map.go` line 838 at r4](https://reviewable.io/reviews/cockroachdb/swiss/21#-NraTKffB1K4jwtG2Nsk:-NraTKffB1K4jwtG2Nsl:b-rdx6ca) ([raw file](https://github.com/cockroachdb/swiss/blob/666532607c5ae2a72222e22ed2fd4b8b9dacd34e/map.go#L838)):*
> ```Go
> 		// yield and adjusting for any directory growth that happened during
> 		// the yield call.
> 		i := originalIndex * (1 << (m.globalDepth() - originalGlobalDepth))
> ```

[nit] we could use an `m.adjustBucketIndex(index, globaldepth) -> index` which adjusts an index recorded at a different global depth.

I wonder if there's a potential simplification here if we don't store the index directly, but every bucket's index is something like `b.indexVal >> (32 - b.globalDepth())`. Existing `indexVal`s then don't have to change when the depth changes. Probably not worth the trouble given you already have this working though.
___
*[`map.go` line 1276 at r4](https://reviewable.io/reviews/cockroachdb/swiss/21#-NraPHbIAXariY7P5AnW:-NraPHbIAXariY7P5AnX:b4ld7ri) ([raw file](https://github.com/cockroachdb/swiss/blob/666532607c5ae2a72222e22ed2fd4b8b9dacd34e/map.go#L1276)):*
> ```Go
> 			}
> 
> 			if i == desiredOffset {
> ```

[nit] these three `if`s could be a single `switch`. The last case could be the `default` and we can check for `ctrlDeleted` and panic only under`invariants`

petermattis (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/21#-:-Nraz7XO6g9-rfrJaI1f:b-814g8k)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @RaduBerinde)

___
*[`map.go` line 1276 at r4](https://reviewable.io/reviews/cockroachdb/swiss/21#-NraPHbIAXariY7P5AnW:-NraybZH5N64Rojh-bxH:b-a1mvh8) ([raw file](https://github.com/cockroachdb/swiss/blob/666532607c5ae2a72222e22ed2fd4b8b9dacd34e/map.go#L1276)):*

I was suggesting to change `case targetGroup.ctrls.Get(target) == ctrlDeleted` with `default:` and do an invariants-only assertion insine that it is `ctrlDeleted` which it must be because we matched on empty or deleted. Either way is fine with me though.
</blockquote></details>

Ack. I'm going to leave as-is. Rehashing-in-place isn't performance critical enough for the extra comparison to matter.

RaduBerinde:
> Is there anything I'm missing?

I looked some more and AFAICT it's only for iteration as in going through all elements in the map. Their `iterator` can contain just a pointer to the control and a pointer to the slot and it can move those forward until it sees the sentinel.

https://github.com/abseil/abseil-cpp/blob/master/absl/container/internal/raw_hash_set.h#L2053

petermattis:
> I looked some more and AFAICT it's only for iteration as in going through all elements in the map. Their iterator can contain just a pointer to the control and a pointer to the slot and it can move those forward until it sees the sentinel.

Thanks. That is what I saw as well. We use a different technique for iteration so this doesn't apply. I have another PR brewing that will remove the sentinel control byte.

PR #22, opened 2024-02-26
swiss: label benchmark name fields

aclements:
This labels the subbenchmark name fields so they're amenable to the query language used by benchstat and related tools.


petermattis (approved):
LGTM

TIL about these labels. I need to update my version of `benchstat`.

petermattis:
Thanks for the contribution.

PR #23, opened 2024-02-26
swiss: report performance counters in benchmarks

aclements:
This adds a dependency on github.com/aclements/go-perfevent

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,11 +1,15 @@
 module github.com/cockroachdb/swiss
 
-go 1.20
+go 1.21
 
-require github.com/stretchr/testify v1.8.4
+require (
+	github.com/aclements/go-perfevent v0.0.0-20240301234650-f7843625020f
+	github.com/stretchr/testify v1.8.4
+)
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	golang.org/x/sys v0.17.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,6 @@ jobs:
           - ubuntu-latest
           - macos-latest
         go:
-          - '1.20'
           - '1.21'
           - '1.22'
 
--- a/bench_test.go
+++ b/bench_test.go
@@ -19,6 +19,8 @@ import (
 	"io"
 	"strconv"
 	"testing"
+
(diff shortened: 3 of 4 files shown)


go.mod
```diff
@@ -1,11 +1,15 @@
 module github.com/cockroachdb/swiss
 
-go 1.20
+go 1.22.0
```
petermattis, line 3: Did you mean to change this? I don't have a strong need to support older versions of Go, but want to make sure this was intentional. If it was, we'll need to adjust `.github/workflows/ci.yaml` to not test on go1.20 or go1.21.

aclements: I used a few range-over-int loops in go-perfevent, which bumped this version. I took those out so now go-perfevent uses `go 1.20` and this can stay as-is.

aclements:
> Do you know differently?

Unfortunately, that's generally true. Hardware performance counters are quite difficult to virtualize safely. I believe AWS does support hardware PMUs on some instance types, but I haven't tried it: https://www.intel.com/content/www/us/en/developer/articles/technical/intel-vtune-amplifier-functionality-on-aws-instances.html

go.mod
petermattis:
```
Error: ../../../go/pkg/mod/github.com/aclements/go-perfevent@v0.0.0-20240228224405-280aa97c1800/perf/counter.go:235:15: undefined: binary.NativeEndian
Error: ../../../go/pkg/mod/github.com/aclements/go-perfevent@v0.0.0-20240228224405-280aa97c1800/perf/counter.go:240:24: undefined: binary.NativeEndian
Error: ../../../go/pkg/mod/github.com/aclements/go-perfevent@v0.0.0-20240228224405-280aa97c1800/perf/counter.go:241:24: undefined: binary.NativeEndian
Error: ../../../go/pkg/mod/github.com/aclements/go-perfevent@v0.0.0-20240228224405-280aa97c1800/perf/counter.go:245:27: undefined: binary.NativeEndian
```

I'm fine with bumping the minimum version go1.21. That would also allow the use of `min` and `max` which I explicitly avoided so far.

aclements: Ugh. That's what I get for not actually testing with Go 1.20. Pushed an update to go-perfevent to require Go 1.21, and a new patch here to update the dependency and Go version.

petermattis: You're going to have to delete go1.20 from the [.github/workflows/ci.yaml](https://github.com/cockroachdb/swiss/blob/main/.github/workflows/ci.yaml#L17). Should be a one line change.

aclements: Done. (Sorry, I'm not very used to working with GitHub workflows.)

petermattis:
Thanks for the contribution.

PR #24, opened 2024-02-26
swiss: remove sentinel

petermattis:
Remove usage of the the sentinel control byte which was not serving any purpose. Abseil uses the sentinel control byte to terminate iteration (or claims to), but in this implementation iteration is terminated by iterating over all of the groups exactly once. Mentions that probing are terminated by reaching an empty control byte or the sentinel were erroneous. We only terminate probing by reaching an empty slot which implies that the map always needs at least 1 empty slot for correct operation.

### DIFF
--- a/map.go
+++ b/map.go
@@ -31,35 +31,32 @@
 // Swiss tables is the usage of a separate metadata array that stores 1 byte
 // per slot in the table. 7-bits of this "control byte" are taken from
 // hash(key) and the remaining bit is used to indicate whether the slot is
-// empty, full, deleted, or a sentinel. The metadata array allows quick
-// probes. The Google implementation of Swiss tables uses SIMD on x86 CPUs in
-// order to quickly check 16 slots at a time for a match. Neon on arm64 CPUs
-// is apparently too high latency, but the generic version is still able to
-// compare 8 bytes at time through bit tricks (SWAR, SIMD Within A Register).
+// empty, full, or deleted. The metadata array allows quick probes. The Google
+// implementation of Swiss tables uses SIMD on x86 CPUs in order to quickly
+// check 16 slots at a time for a match. Neon on arm64 CPUs is apparently too
+// high latency, but the generic version is still able to compare 8 bytes at
+// time through bit tricks (SWAR, SIMD Within A Register).
 //
 // Google's Swiss Tables layout is N-1 slots where N is a power of 2 and
 // N+groupSize control bytes. The [N:N+groupSize] control bytes mirror the
 // first groupSize control bytes so that probe operations at the end of the
-// control bytes array do not have to perform additional checks. The control
-// byte for slot N is always a sentinel which is considered empty for the
-// purposes of probing but is not available for storing an entry and is also
-// not a deletion tombstone. The separation of control bytes and slots implies
-// 2 cache misses for a large map (larger than L2 cache size) or a cold map.
-// The swiss.Map implementation differs from Google's layout: it groups
-// together 8 control bytes and 8 slots which often results in 1 cache miss
-// for a large or cold map rather than separate accesses for the controls and
-// slots. The mirrored control bytes are no longer needed and and groups no
-// longer start at arbitrary slot index, but only at those that are multiples
-// of 8.
+// control bytes array do not have to perform additional checks. The
+// separation of control bytes and slots implies 2 cache misses for a large
+// map (larger than L2 cache size) or a cold map. The swiss.Map implementation
+// differs from Google's layout: it groups together 8 control bytes and 8
+// slots which often results in 1 cache miss for a large or cold map rather
+// than separate accesses for the controls and slots. The mirrored control
+// bytes are no longer needed and and groups no longer start at arbitrary slot
+// index, but only at those that are multiples of 8.
 //
 // Probing is done by taking the top 57 bits of hash(key)%N as the index into
 // the groups slice and then performing a check of the groupSize control bytes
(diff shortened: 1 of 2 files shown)


RaduBerinde (commented):
> We only terminate probing by reaching an empty slot which implies that the map always needs at least 1 empty slot for correct operation.

By the way, we could write a randomized test for this: fill all slots except a random one, and then try to insert a random element (rinse, repeat).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/24#-:-Nrb2Kcu57wtj3WZtzzR:b-hv2qhb)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @petermattis)

petermattis (commented):
TFTR!

> We can now use all 2^N slots no? We can rename `.capacity` to `.slotMask`, but there are a lot of other cleanups needed around "capacity".

Yes, we can now use all 2^N slots. Renaming `capacity` to `slotMask` is a bit involved, particularly if we start computing `capacity` as `slotMask+1`. I started a commit to do this and then quickly saw how many threads there were to pull. I think this is a good idea, but I'd like to leave a TODO and perform this in a separate PR.

> By the way, we could write a randomized test for this: fill all slots except a random one, and then try to insert a random element (rinse, repeat).

I did add an invariant check that there is at least one empty slot. I might be misunderstanding your suggestion, but if we fill all of the slots except one we shouldn't be able to use that last slot as doing so would break the probing invariant that there is at least one empty slot. We also can't get into that state anyways due to the way `growthLeft` is calculated (we'd resize, rehash, or split first).

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/24#-:-Nrb94Nz4K4bKV18zNFY:b-8bnoi4)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`map.go` line 40 at r1](https://reviewable.io/reviews/cockroachdb/swiss/24#-Nrb1FZBA7tcUuHn9ynd:-NrbA3BM8VeckXZxnW3y:b-2os5nn) ([raw file](https://github.com/cockroachdb/swiss/blob/818727913ea891ae76228f95ebe29b29c50b32ea/map.go#L40)):*

Ah, Abseil is Google's implementation, I thought it's a separate one.
</blockquote></details>

Yep, though it is possible the internal Google implementation has diverged from the Abseil implementation. I'm not sure how or if they are kept in sync.

RaduBerinde (commented):
Yeah a TODO is fine. The one thing that seems to be missing is a `+1` when we calculate `growthLeft` (but that's not much of an issue since we have a max fill factor anyway).

> I did add an invariant check that there is at least one empty slot. I might be misunderstanding your suggestion, but if we fill all of the slots except one we shouldn't be able to use that last slot as doing so would break the probing invariant that there is at least one empty slot. We also can't get into that state anyways due to the way `growthLeft` is calculated (we'd resize, rehash, or split first).

I see, right. What I wanted to test is that all probe sequences finish (which could be a problem if we have a bug in the triangular numbers thing). We could fill up the bucket (just by setting the control bytes) until only one slot is left, and then look up random elements which are not in the map. But I see we already have a test for this at the probeSeq level so it's not necessary.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/24#-:-NrbM5Wq2sNfwtrgIAcx:bm77e7m)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @petermattis)

___
*[`map.go` line 1598 at r3](https://reviewable.io/reviews/cockroachdb/swiss/24#-NrbNnyX6uJWzOnTgzsk:-NrbNnyX6uJWzOnTgzsl:b2y1new) ([raw file](https://github.com/cockroachdb/swiss/blob/1fdfd75a0f722681e30d7104c5d00876cae7cf02/map.go#L1598)):*
> ```Go
> // triangular progression of the form
> //
> //	p(i) := groupSize * (i^2 + i)/2 + hash (mod mask+1)
> ```

I should have noticed this for the previous PR but this description needs to be updated. We can simplify to just say it's a probe sequence that we use at the group level.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/24#-:-NrbPM3V5I4oQVc2OlWu:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @petermattis)

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/24)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`map.go` line 1598 at r3](https://reviewable.io/reviews/cockroachdb/swiss/24#-NrbNnyX6uJWzOnTgzsk:-NrbPftq-m0K-mfxix8c:b-bzhjt4) ([raw file](https://github.com/cockroachdb/swiss/blob/1fdfd75a0f722681e30d7104c5d00876cae7cf02/map.go#L1598)):*

I should have noticed this for the previous PR but this description needs to be updated. We can simplify to just say it's a probe sequence that we use at the group level.
</blockquote></details>

Ah, this is the problem with too much commentary. So easy to miss updates when making changes. Fixed. I also removed the language below about mirrored control bytes which no longer exist.

RaduBerinde (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/24#-:-NrbY71Y6nlEwB1Pm1b4:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved

PR #25, opened 2024-02-27
swiss: silence checkptr errors

petermattis:
Silence the race-build checkptr warnings that are triggering on
`unsafeSlice.At` and `unsafeConvertSlice` when accessing and constructing the
empty group. The warnings are semi-valid as the empty group slice is actually
the wrong size, but we never access `group.slots` on this group so that is ok.

### DIFF
--- a/map.go
+++ b/map.go
@@ -1663,6 +1663,8 @@ func makeUnsafeSlice[T any](s []T) unsafeSlice[T] {
 }
 
 // At returns a pointer to the element at index i.
+//
+//go:nocheckptr
 func (s unsafeSlice[T]) At(i uintptr) *T {
 	var t T
 	return (*T)(unsafe.Add(s.ptr, unsafe.Sizeof(t)*i))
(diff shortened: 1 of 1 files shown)


petermattis:
I'm not entirely sure this is the right thing to do or if there is another way to avoid these checks in race builds.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/25#-:-NrcXJeI-5LmHtwW7zOG:bnfp4nl)* status: 0 of 1 files reviewed, all discussions resolved

petermattis:
TFTR!

I poked around a little more. I could get rid of the `go:nocheckptr` declaration on `unsafeConvertSlice` by not using it to convert `emptyCtrls` to the "emptyGroup" slice. Doesn't really seem worth it, though. I don't see anything that can be done about the `unsafeSlice.At` call on an "emptyGroup". I've added some additional commentary to explain what is going on and why these declarations are necessary.

PR #26, opened 2024-02-27
swiss: pad bucket to 32 bytes on 32-bit systems

petermattis:


RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/26)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @petermattis)

___
*[`map.go` line 242 at r1](https://reviewable.io/reviews/cockroachdb/swiss/26#-NreaxTp-aIO6wG4okPW:-NretxoS-OSN7lu45Dxe:b8q5wmj) ([raw file](https://github.com/cockroachdb/swiss/blob/4798fdd3e35bd3c48f579435eaaa92c284f25ea0/map.go#L242)):*

I could also assert that `sizeof(bucket)` is 24 bytes on 32-bit systems. My thinking is that indexing a power-of-2 size bucket might be cheaper, though I don't actually know if that is true on 32-bit x86. Thoughts?
</blockquote></details>

I would be surprised if this makes things faster. 24 is an extra shift and add, but we're paying for it with spreading data over more cachelines. I would just change the assertion. We don't care that much about 32-bit anyway..

RaduBerinde (approved):
:lgtm: either way

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/26#-:-NreuTEJ6qNJ3Vnrs1KL:b78w4ar)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @petermattis)

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/26)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`map.go` line 242 at r1](https://reviewable.io/reviews/cockroachdb/swiss/26#-NreaxTp-aIO6wG4okPW:-Nrf8aHq4waeuwH9ebh8:b-nbvb0f) ([raw file](https://github.com/cockroachdb/swiss/blob/4798fdd3e35bd3c48f579435eaaa92c284f25ea0/map.go#L242)):*

I would be surprised if this makes things faster. 24 is an extra shift and add, but we're paying for it with spreading data over more cachelines. I would just change the assertion. We don't care that much about 32-bit anyway..
</blockquote></details>

Thanks for the thoughts. Agreed that we don't care much about 32-bit anyway. I've switched to adjusting the assertion.

PR #27, opened 2024-02-27
swiss: rename bucket.capacity to bucket.slotMask

petermattis:
Now that the sentinel is gone, a bucket's capacity is 2^N slots, not 2^N-1. Adjust various commentary and calculations to reflect this.


petermattis:
I need to do a double-check that I caught all references to capacity.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/27)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on @petermattis)

___
*[`map.go` line 202 at r3](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrfyR6DAuAumqEwRGjY:-NrfyR6DAuAumqEwRGjZ:bsmmeth) ([raw file](https://github.com/cockroachdb/swiss/blob/24dc8122b478ea12fbf1f10c6f000d7134c7d196/map.go#L202)):*
> ```Go
> 	bitsetDeleted = bitsetLSB * uint64(ctrlDeleted)
> 
> 	minBucketCapacity        uint32 = groupSize
> ```

bucket.init() already enforces this, maybe we can remove
___
*[`map.go` line 203 at r3](https://reviewable.io/reviews/cockroachdb/swiss/27#-Nrfyj5J6IrMn5wGJeJW:-Nrfyj5J6IrMn5wGJeJX:b2pk7ex) ([raw file](https://github.com/cockroachdb/swiss/blob/24dc8122b478ea12fbf1f10c6f000d7134c7d196/map.go#L203)):*
> ```Go
> 
> 	minBucketCapacity        uint32 = groupSize
> 	defaultMaxBucketCapacity uint32 = 4096
> ```

[nit] Let's add a comment explaining what this is, even if it duplicates the comment for the `maxBucketCapacity` field (and/or rename to something like `defaultBucketSplitThreshold`)
___
*[`map.go` line 244 at r3](https://reviewable.io/reviews/cockroachdb/swiss/27#-Nrfxbqk-iMM_x-LaLZ2:-Nrfxbqk-iMM_x-LaLZ3:b-tp02wk) ([raw file](https://github.com/cockroachdb/swiss/blob/24dc8122b478ea12fbf1f10c6f000d7134c7d196/map.go#L244)):*
> ```Go
> 
> 	// The total number slots minus 1 (always 2^N-1). The slotMask is used to
> 	// quickly compute i%N using a bitwise & operation.
> ```

Where are we actually using this for an & operation? We definitely use `groupMask` but I don't think we need this to be a mask..

We should either keep it as capacity and let it be `2^N` or remove it entirely and use `(groupMask + 1) * groupSize` when we want to know the capacity in terms of number of slots (which I don't think we need in any hot path).
___
*[`map.go` line 362 at r3](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrfvjqoBtfWiufcDxYp:-NrfvjqoBtfWiufcDxYq:b4j536l) ([raw file](https://github.com/cockroachdb/swiss/blob/24dc8122b478ea12fbf1f10c6f000d7134c7d196/map.go#L362)):*
> ```Go
> 
> 	if initialCapacity > 0 {
> 		// We consider initialCapacity to be an indication from the caller
> ```

It feels like these calculations should be inside `bucket.init`. We should just pass our initial capacity as a hint.
___
*[`map.go` line 1080 at r3](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrfvVhGBH-Z4-VWfT4I:-NrfvVhGBH-Z4-VWfT4J:bxtbxkr) ([raw file](https://github.com/cockroachdb/swiss/blob/24dc8122b478ea12fbf1f10c6f000d7134c7d196/map.go#L1080)):*
> ```Go
> 
> 	b.slotMask = newCapacity - 1
> 	b.groupMask = (newCapacity+groupSize-1)/groupSize - 1
> ```

we know that newCapacity is a power of two so we shouldn't need `groupSize-2`

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/27)* status: 0 of 3 files reviewed, 6 unresolved discussions (waiting on @RaduBerinde)

___
*[`map.go` line 313 at r1](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrfBX7l9zA8SKjjS5js:-NrfBX7l9zA8SKjjS5jt:b61hxun) ([raw file](https://github.com/cockroachdb/swiss/blob/bbcd32b31626fe333a177e906ed4ba0d303014f0/map.go#L313)):*
> ```Go
> 		return v
> 	}
> 	return uint32(1) << 31
> ```

This was necessary in order to handle `WithMaxBucketCapacity(math.MaxUint32)`. I should add a comment to that effect.
___
*[`map.go` line 202 at r3](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrfyR6DAuAumqEwRGjY:-Nrg5tvx8v7w8mF0_p8_:bve6a4f) ([raw file](https://github.com/cockroachdb/swiss/blob/24dc8122b478ea12fbf1f10c6f000d7134c7d196/map.go#L202)):*

bucket.init() already enforces this, maybe we can remove
</blockquote></details>

Sort of. We divide by `Map.maxBucketCapacity` in `Map.Init` so this is useful in avoiding a divide by zero. I have gone ahead and removed this constant and just use `groupSize` instead during initialization.
___
*[`map.go` line 203 at r3](https://reviewable.io/reviews/cockroachdb/swiss/27#-Nrfyj5J6IrMn5wGJeJW:-Nrg64nX06IsO1tOivej:ba4lxip) ([raw file](https://github.com/cockroachdb/swiss/blob/24dc8122b478ea12fbf1f10c6f000d7134c7d196/map.go#L203)):*

[nit] Let's add a comment explaining what this is, even if it duplicates the comment for the `maxBucketCapacity` field (and/or rename to something like `defaultBucketSplitThreshold`)
</blockquote></details>

I've added a comment explaining what this constant does, but left the name the same.
___
*[`map.go` line 244 at r3](https://reviewable.io/reviews/cockroachdb/swiss/27#-Nrfxbqk-iMM_x-LaLZ2:-Nrg6R6eEvAzsWINwsCS:b-7f264p) ([raw file](https://github.com/cockroachdb/swiss/blob/24dc8122b478ea12fbf1f10c6f000d7134c7d196/map.go#L244)):*

Where are we actually using this for an & operation? We definitely use `groupMask` but I don't think we need this to be a mask..

We should either keep it as capacity and let it be `2^N` or remove it entirely and use `(groupMask + 1) * groupSize` when we want to know the capacity in terms of number of slots (which I don't think we need in any hot path).
</blockquote></details>

Ah, good point that we never `&m.slotMask` anywhere. Renamed this field back to `capacity`. I agree that we could get rid of this field now because it isn't used on any hot path.
___
*[`map.go` line 362 at r3](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrfvjqoBtfWiufcDxYp:-Nrg78Ii0MyipqdjZOBK:b-lrmfzz) ([raw file](https://github.com/cockroachdb/swiss/blob/24dc8122b478ea12fbf1f10c6f000d7134c7d196/map.go#L362)):*

It feels like these calculations should be inside `bucket.init`. We should just pass our initial capacity as a hint.
</blockquote></details>

These calculations are only needed when initially sizing a map, though. I wouldn't want to move the calculation of how big to make the directory into `bucket.init`. Am I misunderstanding what you're suggesting?
___
*[`map.go` line 1080 at r3](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrfvVhGBH-Z4-VWfT4I:-Nrg76SSAzrdrqygi7MO:b-qu6n94) ([raw file](https://github.com/cockroachdb/swiss/blob/24dc8122b478ea12fbf1f10c6f000d7134c7d196/map.go#L1080)):*

we know that newCapacity is a power of two so we shouldn't need `groupSize-2`
</blockquote></details>

Good point. Done.

RaduBerinde (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/27#-:-NrgJ9LZ1wDwW2ylyEAo:bnfp4nl)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @petermattis)

___
*[`map.go` line 362 at r3](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrfvjqoBtfWiufcDxYp:-NrgCDbz8xShy1EXZQK6:be6lr0z) ([raw file](https://github.com/cockroachdb/swiss/blob/24dc8122b478ea12fbf1f10c6f000d7134c7d196/map.go#L362)):*

These calculations are only needed when initially sizing a map, though. I wouldn't want to move the calculation of how big to make the directory into `bucket.init`. Am I misunderstanding what you're suggesting?
</blockquote></details>

Ah, yes, don't mind me.
___
*[`map.go` line 244 at r5](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrgHSO1B2N8-uxexk-n:-NrgHSO1B2N8-uxexk-o:b-roek13) ([raw file](https://github.com/cockroachdb/swiss/blob/5efde261b9ee5bc40110ac6a88fd4c7815940a9c/map.go#L244)):*
> ```Go
> 	// groupMask fields.
> 
> 	// The total number (always 2^N). Equal to `(groupMask+1)*groupSize`.
> ```

(unless the bucket is empty, when it is 0)
___
*[`map.go` line 1114 at r5](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrgHJD0CvvXDbYgPmWV:-NrgHJD0CvvXDbYgPmWW:b-xlngdw) ([raw file](https://github.com/cockroachdb/swiss/blob/5efde261b9ee5bc40110ac6a88fd4c7815940a9c/map.go#L1114)):*
> ```Go
> 	b = m.installBucket(b)
> 
> 	if oldCapacity > 0 {
> ```

[nit] we could move this up and put the `for` loop inside it

petermattis (commented):
TFTR!

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/27#-:-Nrgbhf7EE3ASUkS2D7z:b-814g8k)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @RaduBerinde)

___
*[`map.go` line 244 at r5](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrgHSO1B2N8-uxexk-n:-NrgbOxC6bgLAYHP4bHw:b-896fix) ([raw file](https://github.com/cockroachdb/swiss/blob/5efde261b9ee5bc40110ac6a88fd4c7815940a9c/map.go#L244)):*

(unless the bucket is empty, when it is 0)
</blockquote></details>

Done.
___
*[`map.go` line 1114 at r5](https://reviewable.io/reviews/cockroachdb/swiss/27#-NrgHJD0CvvXDbYgPmWV:-Nrgb_LP2tFNLe7IEh97:bcvjgzp) ([raw file](https://github.com/cockroachdb/swiss/blob/5efde261b9ee5bc40110ac6a88fd4c7815940a9c/map.go#L1114)):*

[nit] we could move this up and put the `for` loop inside it
</blockquote></details>

Good idea. Done.

PR #28, opened 2024-02-27
swiss: optimize bucket init by setting control bytes simultaneously

thepudds:
This is a small optimization to set all the controls bytes in a group to empty in a single assignment when initializing a bucket or when using Clear, rather than doing so in a loop.

This saves around 2% when growing an initially empty map, at least at these sizes:

```
goos: linux
goarch: amd64
pkg: github.com/cockroachdb/swiss
cpu: Intel(R) Xeon(R) CPU @ 2.30GHz

                                   │   old       │              new                   │
                                   │   sec/op    │   sec/op     vs base               │
MapPutGrow/swissMap/Int32/6-16       205.8n ± 0%   200.4n ± 1%  -2.65% (p=0.000 n=10)
MapPutGrow/swissMap/Int32/12-16      495.4n ± 0%   481.7n ± 0%  -2.77% (p=0.000 n=10)
MapPutGrow/swissMap/Int32/18-16      948.5n ± 1%   910.0n ± 1%  -4.06% (p=0.000 n=10)
MapPutGrow/swissMap/Int32/24-16      1.067µ ± 1%   1.036µ ± 0%  -2.91% (p=0.000 n=10)
MapPutGrow/swissMap/Int32/30-16      1.847µ ± 1%   1.764µ ± 1%  -4.49% (p=0.000 n=10)
MapPutGrow/swissMap/Int32/64-16      3.780µ ± 1%   3.591µ ± 1%  -4.99% (p=0.000 n=10)
MapPutGrow/swissMap/Int32/128-16     7.618µ ± 3%   7.294µ ± 1%  -4.25% (p=0.000 n=10)
MapPutGrow/swissMap/Int32/256-16     14.81µ ± 0%   14.59µ ± 1%  -1.44% (p=0.000 n=10)
MapPutGrow/swissMap/Int32/512-16     28.90µ ± 2%   28.17µ ± 1%  -2.55% (p=0.000 n=10)
MapPutGrow/swissMap/Int32/1024-16    56.89µ ± 0%   55.31µ ± 1%  -2.79% (p=0.000 n=10)
MapPutGrow/swissMap/Int32/2048-16    111.8µ ± 1%   110.4µ ± 1%  -1.28% (p=0.000 n=10)
MapPutGrow/swissMap/Int32/4096-16    273.2µ ± 1%   275.5µ ± 1%  +0.82% (p=0.029 n=10)
[...]

PR #29, opened 2024-02-27
swiss: optimize bit clear

thepudds:
When clearing a bit in our bitset in the inner loop of Get, Put, and Delete, we don't need to pass the index of the least significant bit we just found -- we can instead just clear the least significant bit.

Using the idiom of `b = b & (b - 1)` is simpler and faster than the current bit clearing code. This compiles down to a `BLSRQ` when building with `GOAMD64=v3`. For `GOARM=7`, it looks like it compiles to a `LEAQ -1(AX), CX` followed by `ANDQ CX, AX`.

Running the benchmarks seemed to show a geomean improvement of -1.80% sec/op on a GCE n1-standard-16 (with `GOAMD64=v3` for both old and new), but it looked a bit noisy so I don't currently entirely trust that number.

### DIFF
--- a/map.go
+++ b/map.go
@@ -432,7 +432,7 @@ func (m *Map[K, V]) Put(key K, value V) {
 				b.checkInvariants(m)
 				return
 			}
-			match = match.remove(i)
+			match = match.removeFirst()
 		}
 
 		match = g.ctrls.matchEmpty()
@@ -554,7 +554,7 @@ func (m *Map[K, V]) Get(key K) (value V, ok bool) {
 			if key == slot.key {
 				return slot.value, true
 			}
-			match = match.remove(i)
+			match = match.removeFirst()
 		}
 
 		match = g.ctrls.matchEmpty()
@@ -603,7 +603,7 @@ func (m *Map[K, V]) Delete(key K) {
 				b.checkInvariants(m)
 				return
 			}
-			match = match.remove(i)
+			match = match.removeFirst()
 		}
 
 		match = g.ctrls.matchEmpty()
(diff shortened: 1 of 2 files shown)


thepudds:
> geomean improvement of -1.80% sec/op [...] but it looked a bit noisy so I don't currently entirely trust that number.

Hi @petermattis, I can re-run the benchmarks maybe with a higher `-count` in the hopes of a less noisy result, but it takes a while to do that, so I was first curious if you are interested in targeted optimizations like this and my earlier #28. 

No worries of course if you are not ready for contributions from random people.

To introduce myself briefly -- I was one of the people commenting on golang/go#54766, and I had an earlier SwissTable prototype implementation. These two PRs are things I had noticed in my own prototype. (I had mostly set my version aside when I had concluded I would need to do some more experimentations with different layouts like placing the control bytes with the KVs in each group to perhaps better align with performance expectations people might have built up based on the current runtime map layout... which is part of what was prompting me to ask those questions in https://github.com/golang/go/issues/54766#issuecomment-1951384322 last weekend).

Finally, thank you for taking the time to comment your implementation so clearly, and I'm more than thrilled to see your excellent work here looks like it might be en route to the Go runtime!

petermattis (approved):
:lgtm:

Thanks for the contributions. I'm good with targeted optimization PRs like this. Ack that running with a sufficiently high `-count` to eliminate noise is onerous.

> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/29#-:-Nrg2TiQ1dIbDEyGlFrk:b-c4vwbd)* status: 0 of 2 files reviewed, all discussions resolved

petermattis:
Thanks for the contribution.

PR #30, opened 2024-02-27
swiss: fail to compile when bucket size changes

zephyrtronium:
The size of map buckets is controlled to remain cache-friendly. The size was checked and would panic during program initialization if it was different from expected, but that requires actually running a program. This change ties the size check to an expression that fails to compile if the expectation isn't matched, so all we need to do is build.

### DIFF
--- a/map.go
+++ b/map.go
@@ -202,16 +202,14 @@ const (
 	// The default maximum capacity a bucket is allowed to grow to before it
 	// will be split.
 	defaultMaxBucketCapacity uint32 = 4096
+
+	expectedBucketSize = ptrSize + 6*4
 )
 
-func _() {
-	// Don't add fields to the bucket unnecessarily. It is packed for
-	// efficiency so that we can fit 2 buckets into a 64-byte cache line on
-	// 64-bit architectures.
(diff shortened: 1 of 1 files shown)


map.go
```diff
@@ -204,14 +204,13 @@ const (
 	defaultMaxBucketCapacity uint32 = 4096
 )
 
-func init() {
+func _() {
```
petermattis, line 207: I think we could get rid of the function now that we're not requiring code to run. That is, the variable declaration below could be defined at the file-level:

```
const expectedBucketSize = ptrSize + 6*4
var _ [0]struct{} = [unsafe.Sizeof(bucket[int, int]{}) - expectedSize]struct{}{}
```

petermattis (approved):
LGTM

I was wondering if something like this was possible. Thanks for showing me how.

petermattis:
Hmm, an invariant check failed in the 32-bit config: https://github.com/cockroachdb/swiss/actions/runs/8080872903/job/22078394346?pr=30. Hard to imagine that is due to this PR. I'll see if that can be reproduced locally.

petermattis:
Thanks for the contribution.

PR #32, opened 2024-02-29
swiss: fix correctness on 32-bit platforms

petermattis:
On 32-bit platforms `uintptr` is 32-bits. We need to compute `globalShift` as
`32 - globalDepth`, while previously we were hardcoding 64.

Simplify the determination for when split failed to move any records or moved
all of the records to the new bucket. The previous conditionals were overly
complex and also slightly erroneous in edge cases where no records were moved,
yet there were tombstones present in the bucket being split.

### DIFF
--- a/map.go
+++ b/map.go
@@ -203,6 +203,14 @@ const (
 	// will be split.
 	defaultMaxBucketCapacity uint32 = 4096
 
+	// ptrSize and shiftMask are used to optimize code generation for
+	// Map.bucket(), Map.bucketCount(), and bucketStep(). This technique was
+	// lifted from the Go runtime's runtime/map.go:bucketShift() routine. Note
+	// that ptrSize will be either 4 on 32-bit archs or 8 on 64-bit archs.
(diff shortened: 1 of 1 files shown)


petermattis:
The hardcoding of 64 rather than using 32 on 32-bit platforms was the cause of this failure: https://github.com/cockroachdb/swiss/actions/runs/8080872903/job/22078394346?pr=30

RaduBerinde (approved):
Sorry for suggesting those conditionals. I think what I was going for was to require each bucket to be under max usage ratio so that we can absorb a bunch of inserts before splitting again. But this is fine, it shouldn't happen in practice.

petermattis:
TFTR!

> Sorry for suggesting those conditionals. I think what I was going for was to require each bucket to be under max usage ratio so that we can absorb a bunch of inserts before splitting again. But this is fine, it shouldn't happen in practice.

Well, I thought they were a good idea too. I was perplexed why the code was deciding to commit to splitting a bucket when we accidentally had a degenerate hash function on 32-bit platforms (due to the miscalculations of `globalShift`).

PR #44, opened 2025-03-03
move ctrlGroup.Get/Set to endian_*.go

RaduBerinde:
I have noticed that `if <constant>` blocks which get omitted can still
affect inlining decisions (they contribute to a complexity estimation
that happens before the block is omitted). Move the
`ctrlGroup.Get/Set` methods to the tag-dependent files and remove the
recently added invariant-only bound check.

### DIFF
--- a/endian_big.go
+++ b/endian_big.go
@@ -17,4 +17,18 @@
 
 package swiss
 
+import "unsafe"
+
 const bigEndian = true
+
+// Get returns the i-th control byte.
+func (g *ctrlGroup) Get(i uint32) ctrl {
+	i = i ^ (groupSize - 1) // equivalent to (groupSize-1-i).
+	return *(*ctrl)(unsafe.Add(unsafe.Pointer(g), i))
+}
+
+// Set sets the i-th control byte.
+func (g *ctrlGroup) Set(i uint32, c ctrl) {
+	i = i ^ (groupSize - 1) // equivalent to (groupSize-1-i).
+	*(*ctrl)(unsafe.Add(unsafe.Pointer(g), i)) = c
+}
--- a/endian_little.go
+++ b/endian_little.go
@@ -17,4 +17,16 @@
 
 package swiss
 
+import "unsafe"
+
 const bigEndian = false
+
+// Get returns the i-th control byte.
+func (g *ctrlGroup) Get(i uint32) ctrl {
+	return *(*ctrl)(unsafe.Add(unsafe.Pointer(g), i))
+}
+
+// Set sets the i-th control byte.
+func (g *ctrlGroup) Set(i uint32, c ctrl) {
+	*(*ctrl)(unsafe.Add(unsafe.Pointer(g), i)) = c
+}
--- a/map.go
+++ b/map.go
@@ -1472,29 +1472,9 @@ type ctrl uint8
 
 // ctrlGroup is a fixed size array of groupSize control bytes stored in a
 // uint64.
+//
+// See Get() and Set() methods implemented in endian_*.go.
 type ctrlGroup uint64
-
-// Get returns the i-th control byte.
-func (g *ctrlGroup) Get(i uint32) ctrl {
-	if invariants && i >= groupSize {
-		panic("invalid index")
-	}
-	if bigEndian {
-		i = i ^ (groupSize - 1) // equivalent to (groupSize-1-i).
-	}
-	return *(*ctrl)(unsafe.Add(unsafe.Pointer(g), i))
-}
-
-// Set sets the i-th control byte.
-func (g *ctrlGroup) Set(i uint32, c ctrl) {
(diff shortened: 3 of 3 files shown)


RaduBerinde:
To be clear - I expect these functions to be inlined regardless.. But conceivably it could make a difference as to whether the caller or the caller's caller gets inlined as well.

petermattis (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/44)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[`endian_big.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/swiss/44#-OKSM5j14gHtXdtpKhY9:-OKSM5j14gHtXdtpKhYA:b-uy5n4t) ([raw file](https://github.com/cockroachdb/swiss/blob/3d4f59cec98790ce4e870b727f07f52bae523e96/endian_big.go#L26)):*
> ```Go
> // Get returns the i-th control byte.
> func (g *ctrlGroup) Get(i uint32) ctrl {
> 	i = i ^ (groupSize - 1) // equivalent to (groupSize-1-i).
> ```

Is there a reason this alternative xor formulation is preferable? Seems less readable. Presumably it is faster, but I'm not seeing why.

RaduBerinde (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/44)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @petermattis)

___
*[`endian_big.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/swiss/44#-OKSM5j14gHtXdtpKhY9:-OKSNYR_D7vuLoX-VJN1:bpi3aqj) ([raw file](https://github.com/cockroachdb/swiss/blob/3d4f59cec98790ce4e870b727f07f52bae523e96/endian_big.go#L26)):*

Is there a reason this alternative xor formulation is preferable? Seems less readable. Presumably it is faster, but I'm not seeing why.
</blockquote></details>

I assumed an xor would be faster than a sub..

petermattis (commented):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/44)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`endian_big.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/swiss/44#-OKSM5j14gHtXdtpKhY9:-OKSQxYu9IpkSypBMk-r:b7czhc2) ([raw file](https://github.com/cockroachdb/swiss/blob/3d4f59cec98790ce4e870b727f07f52bae523e96/endian_big.go#L26)):*

I assumed an xor would be faster than a sub..
</blockquote></details>

My mental model is these operations should be about equivalent perf. ChatGPT claims they are usually both 1-cycle instructions: https://chatgpt.com/share/e/67c5ffe8-b6d4-8013-9fb6-6c4be13e515f. Do benchmarks show any difference?

RaduBerinde:
*[`endian_big.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/swiss/44#-OKSM5j14gHtXdtpKhY9:-OKSUZgX99VRjmcootOD:b-x224d6) ([raw file](https://github.com/cockroachdb/swiss/blob/3d4f59cec98790ce4e870b727f07f52bae523e96/endian_big.go#L26)):*

My mental model is these operations should be about equivalent perf. ChatGPT claims they are usually both 1-cycle instructions: https://chatgpt.com/share/e/67c5ffe8-b6d4-8013-9fb6-6c4be13e515f. Do benchmarks show any difference?
</blockquote></details>

You're probably right, I'm probably living in the 90s haha  It's just that I don't know that much about s390x (this is big-endian-only code). I don't have a machine like that to benchmark on. GPT says that both are 1 cycle on modern s390x too. I can change it to `-` if you prefer.

petermattis (approved):
> *[Reviewable](https://reviewable.io/reviews/cockroachdb/swiss/44)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @RaduBerinde)

___
*[`endian_big.go` line 26 at r1](https://reviewable.io/reviews/cockroachdb/swiss/44#-OKSM5j14gHtXdtpKhY9:-OKSh-9b2Guy-HjkOOf8:b-ss65qi) ([raw file](https://github.com/cockroachdb/swiss/blob/3d4f59cec98790ce4e870b727f07f52bae523e96/endian_big.go#L26)):*

You're probably right, I'm probably living in the 90s haha  It's just that I don't know that much about s390x (this is big-endian-only code). I don't have a machine like that to benchmark on. GPT says that both are 1 cycle on modern s390x too. I can change it to `-` if you prefer.
</blockquote></details>

Fine to leave as-is.

RaduBerinde:
TFTR!

PR #46, opened 2025-03-11
swiss: reset hash when map becomes empty

cristaloleg:
Basically this is a fix after the comment on the original issue. https://github.com/golang/go/issues/25237#issuecomment-686921752 

Same line can be found in the new Go 1.24 map https://github.com/golang/go/blob/2d097e363a6fce725802ecbde6d0d1b90f45290d/src/internal/runtime/maps/map.go#L664

### DIFF
--- a/map.go
+++ b/map.go
@@ -593,6 +593,12 @@ func (m *Map[K, V]) Delete(key K) {
 			if key == s.key {
 				b.used--
 				m.used--
+				if m.used == 0 {
+					// Reset the hash seed to make it more difficult for attackers to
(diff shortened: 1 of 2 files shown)


petermattis:
Hi @cristaloleg. Apologies for the delay in reviewing. The change itself looks good. Is there any way we can test this? The test can even be somewhat intrusive on the `Map` internals. For example, delete all of the elements from the `Map` and verify that `Map.seed` changes when the map becomes empty.

cristaloleg:
Test added.

petermattis (approved):
Perfect. Thanks.

